PR #15, opened 2016-07-05
First gh_pages test

ocefpaf:
@jbosch-noaa this is the first test to go from the notebook to HTML via `jekyll`. The theme I am using is very simple and it is far from what we want for IOOS.

@lukecampbell you mentioned you had a jekyll theme that had the same look-and-feel of the IOOS new pages. If you can point me to that I can test it here. If that is not usable we need to tweak the theme until we reach the desirable style.


lukecampbell:
I don't have a jekyll theme, I just have the CSS files from the ioos.us website. I can take a look at updating the CSS stuff next week.

PR #18, opened 2016-08-08
Adding HFR notebooks

jbosch-noaa:
@ocefpaf  -These are my two latest notebooks where I learn about the HFR thredds


notebooks/hfradar/CC_NDBC_SIO_HFRdata.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
```
ocefpaf: We need to find a better way to save the output. Because `stdout` is no good.

rsignell-usgs:
How shall we comment on these notebooks?  
Via new issues, or here? 

Why does lonlat_box have a time on it?
```
SIO_query.lonlat_box(north = 42,south=35,west=-80, east=-69).time(datetime(2015, 7, 17, 23))
```

ocefpaf:
I like new issues for merged stuff. But since this is @jbosch-noaa's baby I will let her decided.

BTW, since it is hard to follow notebooks on GitHub, @rsignell-usgs probably means 
`cell [25]` of `SIPHON_NCSS_HFR_DataAccess_v2.ipynb`.

jbosch-noaa:
Thank you @ocefpaf .  @rsignell-usgs : This is a very preliminary notebook where I am learning how to best use siphon.  The script is my no means complete.  The code was copied from a Unidata script that I got from last summer's Python workshop they hosted in Boulder. I'm not really sure why the time is in there.  It is a little detail (amongst many more) that I planned on asking @dopplershift .  

So Ryan, now that I have dragged you in to the conversation, this notebook is the one we were playing around with at SciPy.

dopplershift:
@rsignell-usgs You're not visually parsing this line correctly:
```python
SIO_query.lonlat_box(north = 42,south=35,west=-80, east=-69).time(datetime(2015, 7, 17, 23))
```
That's chaining together two methods calls on a query object (since each method returns `self` so that we can do things like this). This is a single call to set the lon/lat box and then a call to set the time.

rsignell-usgs:
Ah, I get it.  Thanks @dopplershift !

jbosch-noaa:
@dopplershift  - Thanks for the clarification.  I will probably change that code since I am writing code for beginners of Siphon.  The chained code is not immediately intuitive to novices like me :)

dopplershift:
Absolutely, write whatever works for you. It just made sense to me when creating the API to include that functionality.

ocefpaf:
@jbosch-noaa since you mentioned code clarity here are my two cents.

It may seem clearer to write:

```python
bbox_filtered = SIO_query.lonlat_box(north = 42,south=35,west=-80, east=-69)
time_filtered = bbox_filter.time(datetime(2015, 7, 17, 23))
```

(Disclaimer: this is not tested and I do not know anything about `siphon`!)

However, note that we either have to create two new variables, like above, or we have to clobber the original variable `SIO_query`. Sometimes is can be create code a complexity that chaining methods will remove easily.

Anyways, I try to follow this rule:

- if the chained commands can be read like a phrase: `data.aggregate().sum()` (the chained commands have a single direction) do not split them.
- if the chained commands can be reversed (no specific order) then it is OK to split them \*.

I don't know about `siphon` but I guess that command could be written as:

\* I am assuming that

```python
SIO_query.time(datetime(2015, 7, 17, 23)).lonlat_box(north = 42,south=35,west=-80, east=-69)
```

is equivalent to the original command.

BTW, if that is the case here, it would be better to have an API that would allow:

```python
SIO.query(time=datetime(2015, 7, 17, 23),
          lonlat_box=dict(north=42, south=35, west=-80, east=-69)
)
```

dopplershift:
Just write this:
```python
SIO_query = SIO_ncss.query()
SIO_query.lonlat_box(north = 42,south=35,west=-80, east=-69)
SIO_query.time(datetime(2015, 7, 17, 23))
```
The methods are modifying `SIO_query` in place--no copies. It's just returning `self` as a convenience rather than because things are immutable.

PR #21, opened 2016-08-10
Add HTML output to HFR cc

ocefpaf:
@jbosch-noaa I added the `HTML` output that looks "pretty" on the notebook and that can be saved to a file, same as `JSON` see [here](https://github.com/ioos/compliance-checker/blob/7276abba102e2ef6cb390d49ba0c38726f625913/compliance_checker/runner.py#L43-L50). However it would be nice to save the plain text file too, but that option is not available unless you re-direct the `stdout` to a file in the CLI interface or do some workaround to capture the `stdout` here and save to a file.

Ping @lukecampbell who is the expert on `compliance-checker` :wink:


ocefpaf:
@jbosch-noaa you can view the new notebook [here](http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/a2126dd789ce1deaddbba2b08b9bb44379793797/notebooks/hfradar/CC_NDBC_SIO_HFRdata.ipynb). Unfortunately GitHub is quite bad todiff Jupyter notebooks and the *Files changed* tab is useless here. (That will change when [nbdime](https://github.com/jupyter/nbdime) is released.)

If you are happy with the changes feel free to merge. I opened https://github.com/ioos/compliance-checker/issues/290 to keep track of saving text files.

jbosch-noaa:
Haha .. it's like you saw me go to the files changed tab and thought, huh?  The looks great @ocefpaf .  Thanks.

Thank you for also opening another issue for keeping track of saving text files.

ocefpaf:
In the past we used to commit both the notebook and the code to get a better diff. But that is not enough in most cases due to figures, HTML, or any other rich element the notebook is displaying.

`nbdime` promises to fix that for good. Let's hope so!

PR #33, opened 2016-08-24
Remove outdated badge

ocefpaf:
@rsignell-usgs we should rely only on the badge at the main `README` in the root of the repo.


rsignell-usgs:
Phew, thanks!

PR #35, opened 2016-08-25
Py3k

ocefpaf:
This is ready to go now. I removed `pyhum` from the env b/c we never used that and it is still not py3k compliant. The only missing part is `cc-plugin-glider` which is pending a new release (https://github.com/ioos/cc-plugin-glider/issues/7).


environment.yml
```diff
@@ -36,9 +37,8 @@ dependencies:
     - pandas
     - proj.4
     - pygc
-    - pyhum  # Python 2.7
     - pykdtree
-    - pyoos  # Python 2.7
+    - pyoos
```
ocefpaf: :tada:

environment.yml
```diff
@@ -36,9 +37,8 @@ dependencies:
     - pandas
     - proj.4
     - pygc
-    - pyhum  # Python 2.7
```
ocefpaf: :unamused:

environment.yml
```diff
@@ -3,11 +3,11 @@ channels:
     - conda-forge
     - ioos
 dependencies:
-    - python=2.7
+    - python=3.5
     - anaconda-navigator
     - beautifulsoup4
     - cartopy
-    - cc-plugin-glider  # Python 2.7
+    #- cc-plugin-glider  # Python 2.7
```
ocefpaf: :confused:

ocefpaf:
@rsignell-usgs I would love your input on the `environment.yaml` file update.

PR #38, opened 2016-08-29
Closes #36

### LINKED ISSUE #36: Rename pytools/utilities to ioos_tools
We are no longer using an external module for the tools but we are still naming it badly as @rsignell-usgs pointed it out to me :stuck_out_tongue_winking_eye:

ocefpaf:
Renames the generic `pytools` to `ioos_toos`. Note that this is still a module that we do not want to package and ship! It should live in the scope of this repo.


rsignell-usgs:
This is good we don't package and ship.   It's just one step away from the notebooks this way -- easier for folks to see how "things work".

ocefpaf:
> easier for folks to see how "things work".

That is the goal. If any of those functions becomes essential we should improve the tools (iris, netcdf4-python, pyoos) instead of shipping workarounds.

PR #63, opened 2016-09-06
Added a mean bias (not mean absolute bias) metric

### LINKED ISSUE #56: Use mean error instead of mean absolute error when computing bias
@rsignell-usgs mentioned it would be nice to preserve the sign when computing the bias.

ocefpaf:
@rsignell-usgs if you have some time please take a look at how I am computing the bias to ensure that is OK.

Closes #56


rsignell-usgs:
I finally got around to trying this out, and it works great.  Nice improvement!

PR #70, opened 2016-09-09
Remove index.ipynb

### LINKED ISSUE #67: get rid of index.ipynb?
Do we really want index.ipynb?
Every time we add a notebook, we have to modify the index.ipynb, but if we delete it, folks can just walk the directory structure.

What do folks think?

rsignell-usgs:
Fixes #67


ocefpaf:
We can add a [link](http://nbviewer.jupyter.org/github/ioos/notebooks_demos/tree/master/notebooks/) in the README to the root directory, from `nbviewer`, as an alternative. That makes it easy to navigate the rendered notebooks.

PR #73, opened 2016-09-13
New run with relative dates

### LINKED ISSUE #71: Few model outputs for default dates
I just tried to run the example notebooks for the Boston Light Swim and I only got a couple model outputs in the result:

image

Perhaps the model results for the default dates are no longer available?

ocefpaf:
Closes #71

Note that `today` &pm; 3 days did not find `COASWT4` and `G1SST` did not return any "water" point. That may need some investigation to confirm if everything is OK with those two endpoints and the notebook.


notebooks/boston_light_swim/00-fetch_data.ipynb
```diff
@@ -60,9 +60,11 @@
    "source": [
     "%%writefile config.yaml\n",
     "\n",
+    "# If not specified defaults to today +- 3 days.\n",
+    "# Use the date below to reproduce the last Boston Light Swim event forecast.\n",
     "date:\n",
-    "    start: 2016-8-16 00:00:00\n",
-    "    stop: 2016-8-29 00:00:00\n",
+    "    start: null  # 2016-8-16 00:00:00\n",
+    "    stop: null  # 2016-8-29 00:00:00\n",
```
ocefpaf: @rsignell-usgs this is the main change. The user is free to entry any `start`/`stop` dates. If one of those is undefined the other one will be set to &pm; 3 days. If both are undefined then it will be `today` &pm; 3 days.

Is that OK?

rsignell-usgs: What I was hoping for was the ability to specify EITHER;
- absolute date:
    - start
    - stop

or 

- relative date
    start = -4d
    stop  = 48h

ocefpaf: YAML does not allow for `timedelta` (see https://github.com/yaml/yaml/issues/8), so that means we would need to implement that with some code... I will think about this to see if I can find a safe way to parse all the possible options.

rsignell-usgs: Grrr...   I guess we could just say the units are days to avoid getting too complicated.

rsignell-usgs:
COAWST4 was not updating because geoport-dev was down.   I really need to get those server-checking services running again.

ioos_tools/ioos.py
```diff
@@ -44,16 +44,24 @@ def parse_config(config_file):
...
+            )
+    elif isinstance(start, int) and isinstance(stop, datetime):
+        start = stop - timedelta(days=abs(start))
+    elif isinstance(start, datetime) and isinstance(stop, int):
+        stop = start + timedelta(days=abs(stop))
+    elif isinstance(start, datetime) and isinstance(stop, datetime):
+        pass
+    else:
+        msg = "Expect dates (YYYY-MM-DD hh:mm:ss) or days offest (int).\nGot start={} and stop={}."  # noqa
+        raise ValueError(msg.format(start, stop))
```
ocefpaf: @rsignell-usgs how about now. If I got this if-statement correct the user can mix and match any days offset (`int`) with absolute dates. If both `start` and `stop` are offset they will be relative to `datetime.today()` at midnight.

notebooks/boston_light_swim/00-fetch_data.ipynb
```diff
@@ -60,9 +60,12 @@
...
     "%%writefile config.yaml\n",
     "\n",
+    "# Specify a YYYY-MM-DD hh:mm:ss date or integer day offset.\n",
+    "# If both start and stop are offsets they will be computed relative to datetime.today() at midnight.\n",
+    "# Use the dates commented below to reproduce the last Boston Light Swim event forecast.\n",
     "date:\n",
-    "    start: 2016-8-16 00:00:00\n",
-    "    stop: 2016-8-29 00:00:00\n",
+    "    start: -5 # 2016-8-16 00:00:00\n",
+    "    stop: +4 # 2016-8-29 00:00:00\n",
```
ocefpaf: These are all valid (the &pm; sign is ignored btw):

```yaml
date:
   start: 2016-8-16 00:00:00
   stop: 2016-8-29 00:00:00
```

```yaml
date:
   start: -5
   stop: +4
```

```yaml
date:
   start: 2016-8-16 00:00:00
   stop: 7
```

```yaml
date:
   start: 3
   stop: 2016-8-29 00:00:00
```

notebooks/boston_light_swim/00-fetch_data.ipynb
```diff
@@ -753,8 +756,6 @@
     }
    ],
    "source": [
-    "from ioos_tools.ioos import get_coops_metadata\n",
-    "\n",
```
ocefpaf: This was the only `pep8` infraction I found.

ocefpaf:
We can iterate in another PR if this is not OK. But I am merging this now to avoid conflicts with the PRs I have in line.

Issue #74, opened 2016-09-13
Need to downgrade to python=3.4 in win64


rsignell-usgs:
I needed to downgrade to python=3.4 in the environment.yml to get the IOOS environment working on win64

ocefpaf:
I will build a new `Windows` `Qt` with `jpeg 9*` for IOOS soon to circumvent that. But downgrading is OK for now.

PR #77, opened 2016-09-14
App veyor

ocefpaf:
This will prevent Windows errors like #74 from re-surfacing.


ocefpaf:
The `IOOS` environment is successfully created and loaded on Windows! :tada:

PR #78, opened 2016-09-14
Pep8 notebooks

ocefpaf:
Fix `pep8` infractions.


ocefpaf:
Note to self. I will have to write a script that loops over the notebooks b/c `xargs` + `stdout` is "gluing" the notebooks together and processing them as one leading to false positives.

PR #80, opened 2016-09-15
Try to run notebook appveyor

ocefpaf:
Not sure if `AppVeyor` can handle this test but it would help us find `Linux` specific code and fix it for multi-platform compatibility.


ocefpaf:
Interesting that we got a timeout in Travis-CI but AppVeyor passed  :smile:

PR #84, opened 2016-10-11
Re-add cc-glider-plugin

ocefpaf:
Latest `cc-glider-plugin` supports Python 3. (Also explicitly added `jupyter_client`.)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ dependencies:
     - anaconda-navigator
     - beautifulsoup4
     - cartopy
-    #- cc-plugin-glider  # Python 2.7
+    - cc-plugin-glider
     - cmocean
     - compliance-checker
     - ctd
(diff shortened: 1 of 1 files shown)


ocefpaf:
The failure above is known and related to the https://dev-catalog.ioos.us/csw

See https://travis-ci.org/ioos/notebooks_demos/jobs/166826917#L604

rsignell-usgs:
https://dev-catalog.ioos.us/csw  is working again

rsignell-usgs:
Oh wait, no it's not.   :disappointed:

ocefpaf:
The notebook ran literally a few minutes ago. Let's see if it passes on master. Either way I checked and your example there works with the "stable" catalog. I am preparing a PR to use that one instead.

ocefpaf:
@mwengren do you know if we should still use https://dev-catalog.ioos.us/csw? Or should we move to a stable version.

mwengren:
Uh oh, I have to defer to @lukecampbell here, looks like maybe related to a
similar database issue that happened with CKAN recently.

lukecampbell:
Try again

PR #85, opened 2016-10-11
Fix csw search examples

ocefpaf:
Also added the IOOS conda channel to fix Windows builds until `conda-forge` can starts using `conda-build 2`.


ocefpaf:
I fixed only `csw_search_examples.ipynb` because that is the one I will use in the webpage example. I will take a look into fixing the rest of the notebooks later.

PR #97, opened 2016-10-26
Review changes

### LINKED ISSUE #88: First post
@jbosch-noaa as we spoke I am writing a simpler first post, based on the "fetch data" from System-Test, for the blog.

I am omitting numerical model data and `OWSLib` filters at the moment. This notebook focus on PyOOS collectors and observed data from COOPS-SOS. The reason for that is not only to make the notebook simpler, but also because the `ngdc` catalog is not updated and the `dev-ioos` catalog is unreliable at the moment (see https://github.com/ioos/notebooks_demos/pull/84#issuecomment-253267000). So using any catalog search will render to very few models found and a poor comparison with the observations.

Please let me know what you think. I will polish it tomorrow (better figures and use some QA/QC on the data), and add some text to tell a story.

http://nbviewer.jupyter.org/gist/ocefpaf/571f19177fe725ce5c1a9ae2f4f17f35

PS: @rsignell-usgs do you have any suggestions?

ocefpaf:
Closes #88 

See https://via.hypothes.is/http://nbviewer.jupyter.org/gist/ocefpaf/571f19177fe725ce5c1a9ae2f4f17f35 for the review.


ocefpaf:
Thanks @jbosch-noaa! You can see it at https://ioos.github.io/notebooks_demos/

PS: Due to a mistake on my part another notebook leaked into the page. Fixing it...

PR #107, opened 2016-11-04
Move wiki instructions to a page

### LINKED ISSUE #106: Convert the wiki instructions a page here
It feels weird to click on the install instructions and get re-directed to the wiki. We should make that a first class page.

ocefpaf:
Closes #106

### DIFF
--- a/webpage/conda.md
+++ b/webpage/conda.md
@@ -1,0 +1,143 @@
+---
+title: Installing the IOOS conda environment
+layout: single
+---
+
+For IOOS python users we recommend the free
+[Miniconda](http://conda.pydata.org/miniconda.html) Python distribution,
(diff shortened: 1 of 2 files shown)


webpage/conda.md
```diff
@@ -0,0 +1,143 @@
+---
+title: Installing the IOOS conda environment
+layout: single
+---
```
ocefpaf: @jbosch-noaa this example shows the beauty of Markdown, `Jekyll`, and a single theme. I only copied the wiki, which is already in Markdown and added the "front matter"\*, and it is done! We do not need to worry about about styling, HTML, CSS, nothing.

\* The header that Jekyll needs to understand what type of page is this.

PR #108, opened 2016-11-08
Update 00-fetch_data.ipynb with new CSW endpoints

rsignell-usgs:


ocefpaf:
This PR changes from the dev IOOS catalog to the final stable version and switches `geoport` to `gamone`.

PR #118, opened 2016-11-11
Simpler Dockerfile...

ocefpaf:
... that uses the same env file to avoid forking problems.


lukecampbell:
This PR has my two favorite words combined!

PR #124, opened 2016-11-16
CF/UGRID/SGRID conventions

ocefpaf:
@jbosch-noaa this one is read for review. Here is the `hypothes.is` link: https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/CF-UGRID-SGRID-conventions/notebooks/2016-11-16-CF-UGRID-SGRID-conventions.ipynb


jbosch-noaa:
OK, I made a few grammatical adjustments and clarifications.

jbosch-noaa:
Ok .... should I squash and merge or re base and merge?  What is the difference if it will be added to the base in the end anyway?

ocefpaf:
> Ok .... should I squash and merge or re base and merge? What is the difference if it will be added to the base in the end anyway?

Squash will make all those individual commits into one. rebase merge will add them as-is. I do not have a strong opinion on this here. Since it is not code and the individual commits are not pollution the history we can avoid squashing.

jbosch-noaa:
so I just ran this on my home machine and got this error on line 18:
C:\Users\jennifer.bosch\AppData\Local\Continuum\Miniconda2\envs\ioos\lib\site-packages\numpy\ma\core.py:882: RuntimeWarning: invalid value encountered in less
  return umath.less(x, self.critical_value)

I also got this other error on line 23:
C:\Users\jennifer.bosch\AppData\Local\Continuum\Miniconda2\envs\ioos\lib\site-packages\cartopy\io\__init__.py:264: DownloadWarning: Downloading: http://naciscdn.org/naturalearth/10m/physical/ne_10m_coastline.zip
  warnings.warn('Downloading: {}'.format(url), DownloadWarning)

Should we be concerned?

ocefpaf:
Some software abuse warnings and scare users. The former is telling you some invalid values were found in that operation, but they are expected. The latter is even less important because it is only letting you know that the data was not cached and will be downloaded. If you run that again the second warning will disappear BTW.

jbosch-noaa:
Ok.  Awesome. Thanks for clarifying ... learning!!!

ocefpaf:
The first thing to always check is `error` vs `warning`. The former you need to call :ambulance: :fire: :fire_engine: but the latter is only a :sun_behind_large_cloud:

Issue #4, opened 2016-05-11
Demonstrate accessing and using HFR data


jbosch-noaa:
You may have a script that does this already but if you point me to it I will start playing with it.

ocefpaf:
We have two notebooks on the topic added in https://github.com/ioos/notebooks_demos/pull/18: `CC_NDBC_SIO_HFRdata.ipynb` and `SIPHON_NCSS_HFR_DataAccess_v2.ipynb`. The former is focused on `compliance-checker` and the latter demonstrates how to get HFRadar using siphon, but it is incomplete.

We still need to
- [ ] add more examples,
- [ ] and to finish  `SIPHON_NCSS_HFR_DataAccess_v2.ipynb`.

jbosch-noaa:
Thank you for listing that out @ocefpaf. 

The compliance-checker one will need updating at some point as a more generic notebook (rather than a notebook pointing out flaws in our HFR THREDDS servers).  

I would like to build out the SIPHON notebook to have a plot or two of some data.  We should think of framing it around a science/practical use question.

ocefpaf:
Thanks @jbosch-noaa I am working on `SIPHON_NCSS_HFR_DataAccess_v2.ipynb` now but I updated the list with your comments so I do not lose trackof them.

ocefpaf:
> The compliance-checker one will need updating at some point as a more generic notebook (rather than a notebook pointing out flaws in our HFR THREDDS servers).

I am thinking about how to address this here. `compliance-checker` is designed to be a `CLI` tool and we already [faced a few limitations](https://github.com/ioos/compliance-checker/issues/290) b/c of that design. So maybe we should:

- demonstrate the CLI via direct calls in the notebook,
- and show passing and failing files examples.

jbosch-noaa:
@ocefpaf  - Sorry I have been slow to get back to you on this.  I am preparing for an unrelated workshop next week.  I think that explaining/demonstrating how a CLI based tool like the compliance-checker would be helpful to our blog audience (beginner to intermediate tech savvy scientists) .

When I return from this workshop I will start a WordPress file for the Access Data part of our website that will link to the User Blog that we are developing.

Issue #94, opened 2016-10-19
IOOS Catalog (CWS) Search Example notebook


jbosch-noaa:
@ocefpaf  - I like the idea of this [notebook](https://github.com/ioos/notebooks_demos/blob/master/notebooks/csw_search_example.ipynb) on searching a CSW.  With a few more comments I think it would be a good one to post relatively soon.  I would like to time the post to the release of the new IOOS Catalog.  This way we can use that CSW instead of the NOAA catalog and we can drum up new users.

I'm sure @mwengren would be a good reviewer once we have it all written.

rsignell-usgs:
Cool!   I'm still unsure which endpoint that notebook should be pointing to.  Is it 
`https://data.ioos.us/csw`, `ioos_dev='https://dev-catalog.ioos.us/csw`, or both?

jbosch-noaa:
@rsignell-usgs - Once the IOOS Catalog is released, https://data.ioos.us/csw will be the link.  

Am I correct @mwengren ?

mwengren:
Yes, for now it should be dev-catalog.ioos.us, but sometime in Nov (4 or
later) it will be data.ioos.us.

ocefpaf:
> Yes, for now it should be dev-catalog.ioos.us, but sometime in Nov (4 or
later) it will be data.ioos.us.

https://dev-catalog.ioos.us/csw seems to be down again though:

![cat](https://cloud.githubusercontent.com/assets/950575/19559155/da62ee2e-96a4-11e6-9636-26cfc5fa36c9.png)

Cannot wait for  Nov 4th :wink:

ocefpaf:
PR https://github.com/ioos/notebooks_demos/pull/128 addresses this issue.

jbosch-noaa:
@ocefpaf  - I know we have been trying to debug #128  but what happened to working with this[ notebook](https://github.com/ioos/notebooks_demos/blob/master/notebooks/staged_noteooks/csw_search_example.ipynb)?  We can continue to work with Scripps and NDBC to improve their TDS but in the mean time I think we should post a more simple CSW example now that the new catalog is live.

ocefpaf:
Done in https://github.com/ioos/notebooks_demos/pull/154 and we will have two more soon.

Issue #48, opened 2016-08-31
Get the HFR endpoints included in the ioos.us catalog


rsignell-usgs:
In this [notebook](https://github.com/ioos/notebooks_demos/blob/master/notebooks/hfradar/finding_HFRadar_currents.ipynb) we don't find any current HFR data in the geoportal catalog because the metadata is [extra stale](http://www.ngdc.noaa.gov/geoportal/rest/document?id=%7B041E7399-5C99-4437-9A26-EC891C40AFF8%7D), and we don't find any curent HFR data in the ioos.us dev catalog because the metadata isn't there.   

I assume this is because geoportal was directly harvesting from the thredds catalogs for HFR data, right? 

So @mwengren, if we want this enabled, is the protocol to bark up the tree of SCOOS so they can add it to their WAF, so their WAF can be gobbled up by the master WAF, and ingested nightly into the dev catalog?

mwengren:
@rsignell-usgs we're working up to releasing the harvest registry - which will be the entry point for data to get into dev-catalog.ioos.us, but it's still in beta mode right now. 

To not put the chicken before the egg, the first step to make this happen is for SCCOOS to generate a WAF of current ISO metadata via ncISO and THREDDS (or other means as appropriate).  I am planning to reach out to the RAs on this in the very near future.  I'll include you on that email and if you are able to lend a hand at all with that process that would be great.  We are working on writing up some documentation with best practices for ncISO/THREDDS, but that is very preliminary so far.  We'd appreciate any contributions to that as well if you want to add to it.

rsignell-usgs:
@mwengren, okay, thanks!  Looking forward to the harvest registry!

ocefpaf:
Here is an updated run that shows this issue:

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2016-11-20-finding_HFRadar_currents.ipynb

ocefpaf:
Closing this so we can track it in #128

PR #128, opened 2016-11-19
CSW search HFRadar

ocefpaf:
~~Not ready for review yet. Just commit in case you want to take a look @jbosch-noaa~~

We have a `SOS` search example with `pyoos`  ([2016-10-12-fetching_data.ipynb](https://github.com/ioos/notebooks_demos/blob/master/notebooks/2016-10-12-fetching_data.ipynb)) and this one will be the `CSW` search example before we put both together in the Boston Light Swim notebook.

Here is the `hypothes.is` link for comments:
https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2017-12-15-finding_HFRadar_currents.ipynb

Closes https://github.com/ioos/notebooks_demos/issues/4


ocefpaf:
Closes #94

@jbosch-noaa this is ready for review.

ocefpaf:
@rsignell-usgs, @mwengren, and @jbosch-noaa I am updating this notebook to use the https://data.ioos.us/csw catalog (cell 4), but I cannot find any references for the schemes available to retrieve the data (cell 7-9).

If you compare the updated notebook with the original you will see in cell 10 that the NGDC catalog did provide this information. (Note also the more descriptive title and key in cell 8 of the original.)

updated: http://nbviewer.ipython.org/20c17b80dc6769ba7f7878b81e4d623a

original: https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2016-11-20-finding_HFRadar_currents.ipynb

Should I just drop the `sniff_link` example?

mwengren:
One issue I know of that looks like it might be related is that the Scripps TDS server for some reason does not output any service links in the ncISO endpoints.  I made a temporary WAF to include some HF Radar data for the Catalog release because we didn't quite get NDBC and Scripps to do it themselves before release date, but I only put the process in place for Scripps.  

Here's an example: http://hfrnet.ucsd.edu/thredds/HFRADAR_USWC_hourly_RTV.html?dataset=HFRNet/USWC/6km/hourly/RTV.  If you click on the 'iso' service, no links.  This is what is in the current Catalog. 

Looking at NDBC now I can see their server includes the actual service endpoints, at least for this dataset: http://sdf.ndbc.noaa.gov/thredds/catalog.html?dataset=hfradar_uswc_6km.  

I should be able to duplicate or replace Scripps with NDBC, but not til tomorrow.  Assuming the other HF Radar datasets are the same, you should get your service links back.  

Did the notebook pull HR Radar data from NDBC originally?  My guess is the Scripps server issue has been there for some time.

mwengren:
Looking more closely at your original notebook, it looks like it was connecting to Scripps.  So, if anyone knows why their ncISO on their server isn't returning DAP/WMS/WCS etc service endpoints, please fill me in: 

http://hfrnet.ucsd.edu/thredds/iso/HFR/USWC/6km/hourly/RTV/HFRADAR,_US_West_Coast,_6km_Resolution,_Hourly_RTV_best.ncd

As it is I had to hack the thredds_crawler to get it to work with the Scripps server ncISO service to begin with: https://github.com/mwengren/thredds_crawler/commit/b52c3906906d40a32fdd3f7eeda63e72e9dbc3fa.  THREDDS is not so straightforward to me.

ocefpaf:
Thanks for taking a look @mwengren!

@rsignell-usgs is the catalog/iso expert. Can you see anything obvious that we are missing?

mwengren:
Ok, I added metadata from the NDBC THREDDS server to the Catalog.  NDBC doesn't have the same issue with their ncISO service, so I'm able to export ISO metadata including the service endpoints.  So now there is some duplication of HF Radar services from both Scripps and NDBC, but hopefully you can filter only the useful datasets that actually have service links. 

I tried to run the 'finding_HFRadar_currents' notebook locally, but got an import error on 'ioos_tools' and stopped there.  Is this new?

rsignell-usgs:
I'm not sure why the ISO records from the SIO HFR thredds server do not contain the data endpoints.  The THREDDS and ncISO versions there are pretty old, and old versions of ncISO had some non CF specific assumptions that caused them to fail for some CF-compliant datasets. 

@motero-sio, any chance you could update to the latest 4.6 release of THREDDS? 
This should be as easy as:

sudo su - tomcat
cd <tomcat>/webapps
../bin/shutdown.sh
mv thredds.war thredds.old
rm -rf thredds
wget ftp://ftp.unidata.ucar.edu/pub/thredds/4.6/4.6.6/thredds.war
../bin/startup.sh


-Rich

motero-sio:
I'll talk to Tom and see what we can do.

jbosch-noaa:
@ocefpaf - Can we use just the NDBC THREDDS in the notebook example while we wait for SIO to update/fix their THREDDS?

ocefpaf:
Using the `NGDC` catalog we cannot filter using `dates`, see my example from before, are you sure we want to go that route? I will need to re-write this notebook from "near real-time currents" to "currents" though.

jbosch-noaa:
I think you meant to say NDBC not NGDC (which is now part of NCEI).  The NDBC THREDDS is near real-time.  It just has a rolling list of the past 2 weeks of HFR data.

ocefpaf:
> I think you meant to say NDBC not NGDC (which is now part of NCEI). The NDBC THREDDS is near real-time. It just has a rolling list of the past 2 weeks of HFR data

Nope I meant NGDC catalog that returns the endpoints.

So you mean by-pass the catalog search and hard-code a NDBC THREDDS endpoint?
I can do that but than this would not be a csw catalog search.

I will retry the filter today and see if I can find a NDBC endpoint instead.

ocefpaf:
> I tried to run the 'finding_HFRadar_currents' notebook locally, but got an import error on 'ioos_tools' and stopped there. Is this new?

The notebook is not standalone. You need to either close this PR and run from there or to copy the `ioos_tools` directory and modify the its entry in the path at the first cell.

I just tried to run the notebook again and I could not find the NDBC using dates in the filter (cell 5).
Later on, cell 9, I see that the dates are actually strings instead of the convention `seconds/hours/days from x`. In cell 14 I remove the dates from the filter and I can find the NDBC data, but that really beats the purpose of "finding near real-time" data :-(

http://nbviewer.jupyter.org/gist/ocefpaf/301d0cd06d15389df2808f8a2fc971a6

PS: note that I can sniff the links but OPeNDAP is not easily recognizable (cell 16).

mwengren:
The stumbling block that is preventing the Scripps HFR THREDDS server from being added to the new Catalog/CS-W is that I can't figure out how to make the ncISO service return the XML sections that include the OPeNDAP and WCS service links (SV_ServiceIdentification).  

Example dataset; http://hfrnet.ucsd.edu/thredds/HFRADAR_USWC_hourly_RTV.html?dataset=HFRNet/USWC/500m/hourly/RTV
Corresponding XML from ncISO: http://hfrnet.ucsd.edu/thredds/iso/HFRNet/USWC/500m/hourly/RTV?catalog=http%3A%2F%2Fhfrnet.ucsd.edu%2Fthredds%2FHFRADAR_USWC_hourly_RTV.html&dataset=HFRNet%2FUSWC%2F500m%2Fhourly%2FRTV

Does not include SV_ServiceIdentification sections.

If this can be fixed somehow (or we could copy the approach that NGDC/NCEI used to generate their XML from these services), the same service endpoints will show in the Catalog.  Well, at least if the link above is fixed so that they're included they will, since the process I set up will pick them up automatically.  If it requires another processing approach, I recommend looping in @benjwadams from ASA as he will have to put that in place, and replace the temporary WAF I deployed for these in the Registry (I'll be on leave next week).  

ping @amilan17 in case she has any ideas.

ocefpaf:
This one is mostly to show (IMO)  an `OWSLib` limitation in the the returns. That lead us to write the workaround in #137 

I will put together another `CSW` search example that avoids the HF-radar problems today. Hold on...

jbosch-noaa:
@ocefpaf  - Trying to organize this repo via waffle ... where are we on this issue?  The hypothes.is link is broken.

ocefpaf:
@jbosch-noaa I updated the link above ([here](
https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2017-05-04-finding_HFRadar_currents.ipynb) it is for "clicking" convenience).

We have two problems there:

1. The search returns data that have insuficiente metadata to identify were it comes from, and without an access endpoint (cells 6-10);
2. We do know that data from NDBC exist (cells 11-16), but they have bad dates, making it hard to access the data programmatically; (see the hack on cell 12)
3. If we remove the date constraint in the `csw` filter we **can** find the NDBC endpoints. My guess is that this issue is probably related to (2).

PS0: we had #48 to track this but it is better if we keep the discussing on this PR for simplicity.
PS1: as we agreed in the meeting I'll write an HF-Radar example with hard-coded endpoints. We can revisit this once the data is rectified.

jbosch-noaa:
@mwengren  - is this HFR end point problem something we can discuss on our SCCOOS webinar tomorrow?

ocefpaf:
A catalog search with the parameters below yield:
```
Found 3 records.

[CeNCOOS/Models/ROMS/California ROMS/California Coastal Regional Ocean Modeling System (ROMS) Nowcast]: CA_DAS
[HYbrid Coordinate Ocean Model (HYCOM): Global]: hycom_global
[UCSC California Current System ROMS Nowcast 10km]: UCSC
```

So the situation is worse then before when we got the UUID.


Parameters:

```python
from datetime import datetime, timedelta

# Region: West coast.
min_lon, max_lon = -123, -121
min_lat, max_lat = 36, 40

bbox = [min_lon, min_lat, max_lon, max_lat]
crs = 'urn:ogc:def:crs:OGC:1.3:CRS84'

# Temporal range: past week.
now = datetime.utcnow()
start, stop = now - timedelta(days=(7)), now - timedelta(days=(14))

# Surface velocity CF names.
cf_names = ['surface_northward_sea_water_velocity',
            'surface_eastward_sea_water_velocity']
```

mwengren:
@ocefpaf Looking at your most recent [notebook link](https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2017-05-04-finding_HFRadar_currents.ipynb) it looks like you are now getting the SCRIPPS HFRadar datasets in your initial query results, correct?  So a step in the right direction, I think.

It looks to me like the NDBC HFRadar aggregations don't include the ACDD `time_coverage_start` and `time_coverage_end`  attributes that would enable a time extent to be included in the ISO XML (and therefore IOOS Catalog and CS-W service).  

This is the relevant XPath query used in threddsIso looking for those attributes: https://github.com/Unidata/threddsIso/blob/master/src/main/resources/xsl/nciso/UnidataDD2MI.xsl#L83-L88.

Examples of the SCRIPPS and NDBC NCML and corresponding Catalog datasets from the two THREDDS servers showing the difference in source and results:

SCRIPPS NCML: http://hfrnet-tds.ucsd.edu/thredds/ncml/HFR/USWC/1km/hourly/RTV/HFRADAR_US_West_Coast_1km_Resolution_Hourly_RTV_best.ncd?catalog=http%3A%2F%2Fhfrnet-tds.ucsd.edu%2Fthredds%2Fcatalog%2FHFR%2FUSWC%2F1km%2Fhourly%2FRTV%2Fcatalog.html&dataset=HFR%2FUSWC%2F1km%2Fhourly%2FRTV%2FHFRADAR_US_West_Coast_1km_Resolution_Hourly_RTV_best.ncd

Catalog dataset: https://data.ioos.us/dataset/near-real-time-surface-ocean-velocity-u-s-west-coast1-km-resolution2

NDBC NCML: http://sdf.ndbc.noaa.gov/thredds/ncml/hfradar_uswc_1km?catalog=http%3A%2F%2Fsdf.ndbc.noaa.gov%2Fthredds%2Fcatalog.html&dataset=hfradar_uswc_1km

Catalog dataset: https://data.ioos.us/dataset/near-real-time-surface-ocean-velocity-u-s-west-coast1-km-resolution1

If you look at the Coverage tab on each of the Catalog dataset pages, you can see NDBC is missing time extent (and also link to the source ISO XML from each to see where it's missing in there, if you really want). 

I think you knew this already though based on comments above in the thread, but I guess my question is: do we need to use the NDBC HF Radar DAC for this notebook, or can SCRIPPS be used instead?  

If we need to use NDBC, we should reach out the them about the deficiency in their netCDF files - cc'ing @jbosch-noaa and @kbailey-noaa about that action.  

I guess I see this as a data provider issue rather than a Catalog issue at this point.  The metadata being fed into the Catalog is being propagated through to the search results in your notebook.  We just need to work on the data attribution at the DAC level.  Does this sound right?

ocefpaf:
> @ocefpaf Looking at your most recent [notebook link](https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2017-05-04-finding_HFRadar_currents.ipynb) it looks like you are now getting the SCRIPPS HFRadar datasets in your initial query results, correct?  So a step in the right direction, I think.

That is not the latest, that was what we got a up to a few moths ago. The current situation is reported in  my message above. Below is a notebook run right now to illustrate that, see `cell [5]`:

http://nbviewer.jupyter.org/gist/ocefpaf/e2ebe5b43f81f728d8f4b5089c8e798a

> It looks to me like the NDBC HFRadar aggregations don't include the ACDD `time_coverage_start` and `time_coverage_end`  attributes that would enable a time extent to be included in the ISO XML (and therefore IOOS Catalog and CS-W service).

Looks that way b/c we can find the data if we remove the date filter. However, there are still some metadata issues with the dates in those data. I don't really care about that but I thought we could pass along to the data provider. (See `cell [12]` from [this](https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2017-05-04-finding_HFRadar_currents.ipynb) notebook.)

> I think you knew this already though based on comments above in the thread, but I guess my question is: do we need to use the NDBC HF Radar DAC for this notebook, or can SCRIPPS be used instead?

I was instructed to focus on `NDBC` when we started this notebook in last Nov, but I am fine with whatever we can find :smile: 

However, not only `SCRIPPS`  is no longer being found but, when we did find them, the metadata key was the `UUID`, which is not human friendly and prevent us from writing a nice "literated notebook" on how to use the catalog. (See `cell [6]` from [this](https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2017-05-04-finding_HFRadar_currents.ipynb ) notebook.)

> I guess I see this as a data provider issue rather than a Catalog issue at this point. The metadata being fed into the Catalog is being propagated through to the search results in your notebook. We just need to work on the data attribution at the DAC level. Does this sound right?

Not sure to be honest. I am writing notebooks that are at the end point of the data. and the entry point is the catalog, so it is all catalog issues to me :wink: 

But I understand what you mean and I'll try to write something that would give us more information. Your message above helps a lot.

mwengren:
@ocefpaf  Thanks, tt is hard to see which date these notebooks have been run, so I just grabbed the most recent link I saw you posted on the thread and assumed those were the latest results - I'm sure there are some dates buried in the output of some of those cells if you look closely enough, but it might be useful to put a cell at the top that just prints the run date, if only for debugging.

@benjwadams and @ericmbernier I will need to pass this on to you guys as a task to debug in the CKAN db and pycsw db.  I don't have bandwidth to debug any further right now.  

The SCRIPPS HFRadar dataset shows a time coverage up to a few days ago in [CKAN](https://data.ioos.us/dataset/near-real-time-surface-ocean-velocity-u-s-west-coast1-km-resolution2), but doesn't seem to be returned in CS-W query results.  

That could either be due to the query not being formed properly to match the dataset, or to a sync issue between CKAN and pycsw.    Can you add a task to try to run this down.

@ocefpaf In terms of the issue classification, that's more intended for us internally in IOOS.  I just mean to say we need to take the next step to work with data providers to improve metadata and follow through on that, rather than leave things like this languishing in the 'Catalog issue' bin indefinitely.

ocefpaf:
>  I'm sure there are some dates buried in the output of some of those cells if you look closely enough, but it might be useful to put a cell at the top that just prints the run date, if only for debugging.

Nope. I never added them, that notebook is a working in progress that got buried in the noise, doing it now. Should appear in my next commit and I'll update the conical hypothesis link in the first message of this issue.

rsignell-usgs:
I'm going to revisit getting the Scripps HFR thredds server updated with @motero-sio.     That should solve the problem.

rsignell-usgs:
Tom Cook at Scripps got back to me and it turns out they have already updated their thredds server 
http://hfrnet-tds.ucsd.edu/thredds and indeed, the harvested  ISO metadata now contains the service endpoints (WMS, OPeNDAP, etc.)
https://registry.ioos.us/waf/HF%20Radar%20DAC/fefc1bd4f1efd20cec0a69000ec9fc30d166d536.xml

So *that's* not the problem....   I'll investigate more.

ocefpaf:
> So that's not the problem.... I'll investigate more

I'll re-write simpler filters too to help this afternoon.

mwengren:
@rsignell-usgs I think I already tracked the Scripps HFR issue down for the most part.  See my comments above.  

There's either an issue with the query params in the notebook that aren't matching the Scripps HFR metadata records any more (unlikely), or, the Scripps metadata records aren't being copied from the CKAN database (where they clearly exist, with service endpoints) to pycsw database/CS-W service (more likely).  Or, maybe something else I'm not thinking of.  

I'm hoping RPS/ASA @benjwadams or @ericmbernier can investigate the CKAN/pycsw issue at some point.  But aside from that, maybe someone could tweak the notebook query params slightly to see if relaxing the filters allows those records to be found.

ocefpaf:
> But aside from that, maybe someone could tweak the notebook query params slightly to see if relaxing the filters allows those records to be found.

That is what I plan do to this afternoon, but let me get to it now so you guys have more info. Hold on...

ocefpaf:
See http://nbviewer.jupyter.org/gist/ocefpaf/b45877a314468f6c1e66dcf31eee610d

- `cell [7]` tries the "past week" logic we have in the original notebook and fails;
- `cell [8]` removes the date constraints and find the data (also no more UUID for metada :tada: );
- `cell [10]` tries a new date constraint of "last month" and succeeds.

I guess we may be asking too much for near real time we filtering "last week."

rsignell-usgs:
If I ingest the ISO metadata into our local pycsw here at WHOI, it works fine, so somehow that ISO metadata isn't getting into the CSW side of the house in the IOOS catalog:

https://gist.github.com/anonymous/2afbfdf769ca9c05a9026911689c6b58

mwengren:
Can we compare the actual ISO records coming out of the IOOS Catalog CS-W service against what is shown in CKAN?

If you take the [Catalog's 1KM West Coast HFR Dataset](https://data.ioos.us/dataset/near-real-time-surface-ocean-velocity-u-s-west-coast1-km-resolution2) from Scripps, the time coverage is from 2011 to 8/29/17 - click on Coverage tab at the bottom.

The harvested [XML](https://registry.ioos.us/waf/HF%20Radar%20DAC/34c9d1f187b13b2bf45f108f6fcaf88e40b936a6.xml) from the WAF shows the same dates (well, actually to 8/30 because it's been updated since last harvested by the Registry).

Agreed, it seems the CS-W service may not updating reliably in the Catalog.  Since your previous month date range query matched, that probably means it's worked sometime in the last month.... just not the last week.

mwengren:
It may not be 100% bulletproof, but we've been testing the pycsw synchronization in the Catalog recently and things are looking up.   Wanted to report that out before I forget.

We're going to continue a few more tests the next few weeks, but if you want to revisit testing this notebook (and releasing once we're certain on Catalog status), you can move ahead with that.  

So I think in parallel there should be a discussion with NDBC about adding `time_coverage_start` and `time_coverage_end` attributes to these netCDF files (or, alternatively, just use SCRIPPS instead?).  Am I on track with these next steps?

cc'ing @kbailey-noaa on the NDBC provider issue.  See my [earlier comment](https://github.com/ioos/notebooks_demos/pull/128#issuecomment-326043698) on the NDBC netCDF attribution.

ocefpaf:
Thanks @mwengren. My tests did pass and I'll get back to this as soon as I finish #268.

rsignell-usgs:
`nciso` should be calculating the time extent automagically even if no time range is specified explicitly in the netcdf file.   If it's not, it might be because it's an old version of `nciso` where it made the bad assumption that the time coordinate  was always called `time`.

rsignell-usgs:
@mwengren , okay, I think I discovered the real reason the NDBC metadata has no time extents:  it's because the time coordinate information is not CF compliant.   Instead of time being an integer or floating point array with units of `<time units> since <date>`,  time is represented with ISO strings.   This is not CF compliant.    Folks don't often realize that using ISO strings is not CF-compliant, since netcdf-java reads these just fine.    But other tools don't, including the current version of `nciso`.   

If `nciso` was using netcdf-java functions for calculating the geospatial and time extents it would work, but it seems like `nciso` is kind of rolling it's own:
https://github.com/Unidata/threddsIso/blob/master/src/main/java/thredds/server/metadata/util/ThreddsExtentUtil.java#L278-L279

@lesserwhirls, do you agree?

lesserwhirls:
That could possibly do the trick. For sure something to look at once the command line and TDS versions of ncIso are merged into one codebase.

ocefpaf:
@jbosch-noaa as we discussed before I am no longer trying to use the `NDBC`. The notebook is ready for review.

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/finding_HFRadar/notebooks/2017-12-15-finding_HFRadar_currents.ipynb

mwengren:
@rsignell-usgs So, the threddsIso code snippit you linked to above is what ncIso/threddsIso uses internally to generate the `time_coverage_start` and `time_coverage_end` global attributes you see in these NCML Service links?  

Scripps: [HFR US West Coast 2KM NCML Service](http://hfrnet-tds.ucsd.edu/thredds/ncml/HFR/USWC/2km/hourly/RTV/HFRADAR_US_West_Coast_2km_Resolution_Hourly_RTV_best.ncd?catalog=http%3A%2F%2Fhfrnet-tds.ucsd.edu%2Fthredds%2Fcatalog%2FHFR%2FUSWC%2F2km%2Fhourly%2FRTV%2Fcatalog.html&dataset=HFR%2FUSWC%2F2km%2Fhourly%2FRTV%2FHFRADAR_US_West_Coast_2km_Resolution_Hourly_RTV_best.ncd)

NDBC: [HFR US West Coast 2KM NCML Service](http://dods.ndbc.noaa.gov/thredds/ncml/hfradar_uswc_2km?catalog=http%3A%2F%2Fdods.ndbc.noaa.gov%2Fthredds%2Fhfradar.html&dataset=hfradar_uswc_2km)

In the case of NDBC, of course, it is missing.  

Those attribute values are calculated by ncIso/threddsIso, rather than provided somehow in the NCML aggregation in the THREDDS config?  In the case of NDBC, it can't read the ISO strings and fails to generate the time extents.  That would make sense.   I'm fairly green on configuring THREDDS servers, so trying to understand what is actually happening.

It is interesting in the case of the Scripps server (link above), the `geospatial_min|max` attributes are included twice, once as regular global attributes, and once with the `<group name="CFMetadata">` block.  Presumably the former are included somehow in the source data, and the latter are the ones calculated by ncIso?

cc'ing @kbailey-noaa so we know what specific feedback to provide to NDBC when we get to that point with them.  If this is within scope for this FY, maybe we should discuss with them again.  Either way, we should note this issue for reference.

rsignell-usgs:
@mwengren , yes, your analysis agrees with my understanding.

jbosch-noaa:
Ok @ocefpaf  - I reviewed the notebook and made a few comments.  Once those are addressed I think this is good to go!

ocefpaf:
> Ok @ocefpaf  - I reviewed the notebook and made a few comments.  Once those are addressed I think this is good to go!

Done! Thanks @jbosch-noaa!!

ocefpaf:
:tada:

PR #137, opened 2016-11-23
Get all CSW results and parse service URLs with geolinks

ocefpaf:
~~Not read for merging because this changes the behavior of `service_urls`.~~

Here is a notebook example with these two function: https://gist.github.com/ocefpaf/a7b1f1d5a6ab01a8ee7212fe9a0e60e7


ocefpaf:
Merging this but we need to update all the notebooks that use `service_urls` accordingly.

PR #145, opened 2016-11-28
This is my first pull request :D

jbosch-noaa:
I named the file differently - navigation_v2.yml - but I am testing this PR thing.  Can it be merged and/or changed to navigation.yml if we want to change the website?


webpage/_data/navigation_v2.yml
```diff
@@ -0,0 +1,38 @@
...
+    url: /
+  - title: "Installing the IOOS conda environment"
+    url: /conda
+
+# Site navigation links.
+site_nav:
+  - title: Video Tutorials
+    children:
+      - title: "*Python tools*"
+      - title: "* Catalog driven reproducible workflows for ocean-science"
```
ocefpaf: @jbosch-noaa as I suspect this will not render as expected. Even though the `*` is the proper way to get bullets in markdown the navigation section is created following the theme rules and the theme must be changed in order to achieve that. (Ping @mwengren here in case he knows a workaround.)

ocefpaf:
@jbosch-noaa this will not work. Jekyll returns
```
jekyll 3.2.1 | Error:  (webpage/_data/navigation.yml):
did not find expected key while parsing a block mapping at line 35 column 7
```

when we try to render that change.

PR #146, opened 2016-12-05
Update theme

ocefpaf:
This update will change the date from the bottom to the top.


ocefpaf:
@jbosch-noaa this one is safe to merge with a green CI. It will update the theme and add the date at the top using https://github.com/ioos/ioos_jekyll_theme/pull/25

PR #147, opened 2016-12-05
Change name

ocefpaf:
@jbosch-noaa can you please check if this PR reflect your thinking in https://github.com/ioos/notebooks_demos/issues/133#issuecomment-264871249 ?

Below is a rendered version. Ignore the landing page (`Code Gallery`) because I did not commit any notebook in this test. I want to know if the sections above, links I added to them, and the removal of the side bar are OK.

https://ocefpaf.github.io/notebooks_demos/

PS: Also the `Contact us` is only as placeholder.  Not sure what info must go in there... Do you have an example I can copy?

PR #148, opened 2016-12-06
Capitalize

ocefpaf:
Adresses https://github.com/ioos/notebooks_demos/pull/147#issuecomment-265242853

> 1.  Capitalize title of the blog on the landing page: "The IOOS Data Demo Center"

Done.

> 3. Under the Video Tutorials, we should rename the third group "Python Demonstration Videos" 

Done.

> 2.  I feel like we should have the title of this blog static somewhere on all the sub pages also.  May be at the top of the sidebar? Maybe in the blue bar?  Just so folks don't forget what this page is called. This idea is up for debate.

The side bar and the blue top bar are rendered by the theme and we can only add extra entries (topics on the former and sections on the latter). The image show how that title looks like an extra topic in the side bar (the all caps is part of the theme.)

sidebar

and this is how it looks tile when place at the top banner (it behaves also as a link to the main page the same way the logo is):

title

Which one do you prefer @jbosch-noaa?

### DIFF
--- a/webpage/_data/navigation.yml
+++ b/webpage/_data/navigation.yml
@@ -1,5 +1,7 @@
 # Main links links.
 main:
+  - title: The IOOS Data Demo Center
+    url: /
   - title: "Code Gallery"
     url: /code_gallery
   - title: "Video Tutorials"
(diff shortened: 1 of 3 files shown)


jbosch-noaa:
I like it on the top in the blue bar.  :)

ocefpaf:
> I like it on the top in the blue bar. :)

Done in the last commit. Merge at will :smile:

Issue #129, opened 2016-11-21
AppVeyor failure


ocefpaf:
Not sure why AppVeyor is failing with 

```shell

[Running notebook]: C:\projects\notebooks-demos\notebooks\2016-11-15-glider_data_example.ipynb
[Failed]
An error occurred while executing the following cell:
------------------
import iris
iris.FUTURE.netcdf_promote = True
glider = iris.load(url)
print(glider)
------------------
OSError: NetCDF: Invalid argument
```

`Linux` and `OS X` can execute that notebook just fine :unamused:

ocefpaf:
@rsignell-usgs I cannot reproduce that with my old Windows XP VM. If you find some time can you test that notebook on your machine? Maybe that endpoint is blocking `AppVeyor`...

rsignell-usgs:
I'm getting that same error when I try to run that notebook on windows.   When I try `ncdump` in my Anaconda prompt, I get:

```
(ioos3) C:\Users\rsignell\Documents\GitHub>ncdump -h "http://data.ioos.us/thredd
s/dodsC/deployments/rutgers/ru29-20150623T1046/ru29-20150623T1046.nc3.nc"
CURL Error: SSL connect error
curl error details:
ncdump: http://data.ioos.us/thredds/dodsC/deployments/rutgers/ru29-20150623T1046
/ru29-20150623T1046.nc3.nc: NetCDF: I/O failure
```

ocefpaf:
That explains why I can access on my old Windows XP VM. Thanks!

I will check what we can do about it but we probably have a bad mix of SSL version in conda-forge.

rsignell-usgs:
I try to open in netCDF-Java on windows, I get this:
![2016-11-21_9-50-42](https://cloud.githubusercontent.com/assets/1872600/20487083/185fa188-afd0-11e6-9be2-6f91453bcbc5.png)

That doesn't depend on any python!

ocefpaf:
> That doesn't depend on any python!

I know, it is the openssl version used like I mentioned above. Ee can use a more permissive openssl version. I am investigating that... Although I am not sure it is recommended. The error is there for a reason.

ocefpaf:
It is not something that we can fix. I am trying a few workarounds though, but they may not be recommended. Stay tuned...

ocefpaf:
This is fixed via a patch and the next `netcdf` lib will fix this for good.

PR #149, opened 2016-12-07
Just testing, not for merging

ocefpaf:
Nothing to see here. Just checking how the latest changes in conda-forge (mostly https://github.com/conda-forge/hdf5-feedstock/pull/55) will behave in our envs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_install:
   - conda info -a
 
 install:
-  - conda env create -q -f environment.yml
+  - conda env create --file environment.yml
   - source activate IOOS
 
 script:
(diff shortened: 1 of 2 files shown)


ocefpaf:
Latest updates in conda-forge are OK now to create the env on all platforms, but we are still suffering from https://github.com/ioos/notebooks_demos/issues/129

I will test a few older versions of `libnetcdf` and `netcdf4` here to see if that is related to the packages.

ocefpaf:
I am not happy about the workaround pinning here, but let's keep it that way until we have `qt` on `OS X` (so we can drop icu pin) and figure out the `libnetcdf 4.1.1.1` issue.

PR #153, opened 2016-12-12
Python 2 and 3 envs

ocefpaf:
No ready for merging yet!!! Still need to update the CIs and the docs. Adding this just to start a discussion.

@rsignell-usgs suggested that we should rename the `IOOS` env to `IOOS2` (Python 2) and IOOS3 (Python 3). Ideally we would still use IOOS3 as the default everywhere, but people would have the option to use a similar env by downloading the `environment-python-2.yml` file and creating an with with that one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,25 +6,27 @@ matrix:
   fast_finish: true
   include:
     - os: linux
-      env: TEST_TARGET=default
+      env: TEST_TARGET=default ENV_NAME="IOOS3" ENV="environment.yml"
     - os: linux
-      env: TEST_TARGET=notebooks
+      env: TEST_TARGET=notebooks ENV_NAME="IOOS3" ENV="environment.yml"
     - os: linux
-      env: TEST_TARGET=staged_notebooks
+      env: TEST_TARGET=notebooks_python2 ENV_NAME="IOOS2" ENV="environment-python-2.yml"
     - os: linux
-      env: TEST_TARGET=publish
+      env: TEST_TARGET=staged_notebooks ENV_NAME="IOOS3" ENV="environment.yml"
+    - os: linux
+      env: TEST_TARGET=publish ENV_NAME="IOOS3" ENV="environment.yml"
     - os: osx
-      env: TEST_TARGET=notebooks
+      env: TEST_TARGET=notebooks ENV_NAME="IOOS3" ENV="environment.yml"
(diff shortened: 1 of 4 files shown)


jbosch-noaa:
It's a good idea if we knew where were a lot of users of both.  I thought Python 2 was no longer going to be supported in the next year though.

rsignell-usgs:
@jbosch-noaa , my thinking here was that although we want people to use Python 3, there are a *lot* of examples floating around that won't work on Python.    A lot of the notebooks that I have out there, for example.  :cry:   So we just want to avoid the situation where folks just use the `IOOS` environment, and then don't know why things that used to work anymore when they upgrade their IOOS environment (moving say, from python 2 to python 3 without them realizing it).    IMHO we should be more explicit that we are now using Python 3 and then still provide a way folks can create a Python 2 environment for compatibility.

ocefpaf:
I am stuck trying to solve https://github.com/ioos/notebooks_demos/issues/129, and removing our dependency on the `IOOS` channel (that was meant as a workaround for the lack of conda-build 2 packages in conda-forge.)

As soon as I solve those I will update the docs and merge this.

PS: the fact that we cannot install two pythons in the same env is the real problem here. But that is a `conda` limitation and there is nothing we can do. However, if people install both envs, `IOOS2` and `IOOS3`, they will be able to chose/switch them in the Jupyter menu regardless of the env they first started.

jbosch-noaa:
@rsignell-usgs  and @ocefpaf  - Your arguments are valid.  I just wanted to ask the question before I gave you my thumbs up.  I agree that we need to make sure that we flag at the top of a notebook which version of Python that notebook will run in.

ocefpaf:
> I agree that we need to make sure that we flag at the top of a notebook which version of Python that notebook will run in.

That should not be a problem with any of our notebooks. I always use Python 2/3 compatible syntax.  I can add an extra test to enforce that once I finish this PR.

ocefpaf:
I finally fixed the envs and both `Python 2` and `3` are being created successfully and all the notebooks are passing in both!

@rsignell-usgs and @jbosch-noaa I appreciate if you could take a look at the revised instructions. Feel free to edit them in the [wiki](https://github.com/ioos/notebooks_demos/wiki/Installing-Conda-Python-with-the-IOOS-environment) and I will copy and paste them here later.

PR #154, opened 2016-12-19
Another CSW catalog example

ocefpaf:
@jbosch-noaa here is an alternative CSW example that uses the IOOS catalog to search for Sea Surface Temperature in the Northwest coast. Here is a hypothes.is link for the text review:

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/new_csw_example/notebooks/2016-12-19-exploring_csw.ipynb?flush_cache=true


jbosch-noaa:
Ok, I made some edits to the comments and asked a few aesthetic questions.

ocefpaf:
I addressed the review comments. But the comments from `hypothes.is` did not migrated yet. Weird...

I updated the link above and it should reflect the latest version soon. If not here is the static version:

http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/new_csw_example/notebooks/2016-12-19-exploring_csw.ipynb

PS: the wrapping that you requested in complicated because that text is in the output and the HTML rendered takes care of it by adding an horizontal scroll bar to it.

I can investigate a better way on how to handle that in the future but I am not sure if the best solution is on the python side (my notebook), or on the theme side (jekyll).

notebooks/2016-12-19-exploring_csw.ipynb
```diff
@@ -478,16 +478,21 @@
...
     }
    ],
    "source": [
+    "import textwrap\n",
+    "\n",
+    "\n",
     "value = csw.records['opendap.co-ops.nos.noaa.gov-urn_ioos_station_NOAA.NOS.CO-OPS_9439040']\n",
     "\n",
-    "print(value.abstract)"
+    "print('\\n'.join(textwrap.wrap(value.abstract)))"
```
ocefpaf: @jbosch-noaa I tested this in `Jekyll` and it wraps the output nicely in the HTML. However, I am not wrapping the ones with URLs because they became unreadable. I will let `Jekyll` deal with that by adding the horizontal scroll bar.

PR #159, opened 2016-12-22
Trying to embed video

jbosch-noaa:


webpage/_data/navigation_v2.yml
```diff
@@ -0,0 +1,38 @@
+# Main links links.
+main:
```
ocefpaf: This won't work @jbosch-noaa because the theme will not read that file.

BTW, the reason we use GitHub is to avoid versioning files with `v2` :smile:

PR #166, opened 2017-01-06
Add links to the side-bar

ocefpaf:
@jbosch-noaa this is ready for review.

### DIFF
--- a/webpage/_data/navigation.yml
+++ b/webpage/_data/navigation.yml
@@ -13,3 +13,9 @@ site_nav:
     children:
       - title: "Installing the IOOS conda environment"
         url: /other_resources
+      - title: "Australian ocean data network portal user guide"
+        url: https://help.aodn.org.au/aodn-data-tools/user-code-library/
+      - title: "Unidata jupyter notebook gallery"
+        url: http://unidata.github.io/notebook-gallery/
+      - title: "Extracting and enriching OBIS data with R"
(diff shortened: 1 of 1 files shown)


webpage/_data/navigation.yml
```diff
@@ -13,3 +13,9 @@ site_nav:
     children:
       - title: "Installing the IOOS conda environment"
         url: /other_resources
+      - title: "Australian ocean data network portal user guide"
```
ocefpaf: Should I use `AODN` instead of Australian ocean data network to make this shorter?

jbosch-noaa: I'm concerned folks won't know what AODN is. AODN is part of IMOS - another little know acronym for the Integrated Marine Observing System.  I like using the title "IMOS User Code Library" since it's shorter and that is really what the site gives access to.

ocefpaf: Done in https://github.com/ioos/notebooks_demos/pull/166/commits/07dd6e39a804ec1b95c1fd9f5d4e875a1ed0a020

jbosch-noaa: I was talking to folks here further and I think we should title the IMOS link as this " Opening netCDF files - hints from AODN"

ocefpaf: Alright, done and done ;-)

PR #167, opened 2017-01-06
Add images

ocefpaf:
@jbosch-noaa see http://ocefpaf.github.io/notebooks_demos/video_tutorials/ and https://ocefpaf.github.io/notebooks_demos/

PS: the [Data_Demo_Center_middlepic_size770.png](https://ocefpaf.github.io//notebooks_demos/assets/Data_Demo_Center_middlepic_size770.png) is quite big and heavy to load (~ 6.5M). My flaky internet is suffering :grimacing:

Would you be OK if I make it smaller?


jbosch-noaa:
I can reduce the size of the file and send you a smaller one.  What is the max size it should be?

ocefpaf:
> I can reduce the size of the file and send you a smaller one. What is the max size it should be?

I already did it here. I was just asking if it would be OK to do it. (I was not sure if you meant for it to be big.)

PR #168, opened 2017-01-06
Add virtual space between entries

ocefpaf:
@jbosch-noaa this is the change you requested. If someone clicks between the links, in the empty title, they will be directed to a 404 page. That is the side effect of this hack.


jbosch-noaa:
hummmmm.  Maybe we can just number the links?
like this:

title: "1. Installing the IOOS conda environment"
title: " 2. Opening netCDF files - hints from AODN"

ocefpaf:
Let's try that. Hold on...

ocefpaf:
How does that look?

links

PR #130, opened 2016-11-21
First draft for the Boston Light Swim post

ocefpaf:
@rsignell-usgs this is just a compilation of the 3 notebooks into one. I will make some adjustments to make it more appealing for publication in the site. Also, I will not remove the 3-part version from staged notebooks.

PR #170, opened 2017-01-10
Add getrecords2 options to get_csw_records

ocefpaf:
@rsignell-usgs this extra `kwargs` allow is to pass any argument to `get_csw_records2`.


ocefpaf:
No need to wait for the CIs as this is not tested there yet. We will be using this in #130 soon.

PR #171, opened 2017-01-10
Siphon example

ocefpaf:
@jbosch-noaa this is ready for a first review.

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/siphon/notebooks/2016-12-23-siphon-explore-thredds.ipynb

@kwilcox any idea why the line `service = WebMapService(url)` in cell `[14]` takes more than 10 minutes? That is accessing http://pdx.axiomalaska.com/ncWMS/wms


kwilcox:
This is not pointing to the ncWMS integrated into THREDDS. We are pointing the WMS service link to a stand alone ncWMS instance that hosts every single dataset we have. The GetCaps is enormous.

We are still brainstorming how to solve this issue when we move to ncWMS2. In the meantime there isn't much we can do. Can you tell if it is response time or parsing time that takes so long?

ocefpaf:
> We are pointing the WMS service link to a stand alone ncWMS instance that hosts every single dataset we have. The GetCaps is enormous.

That explains it.

> Can you tell if it is response time or parsing time that takes so long?

Not sure.. I haven't checl. But that is not important for the example there. I will remove that part or hardcode a known layer just to finish that notebook.

Thanks @kwilcox!

ocefpaf:
@jbosch-noaa this should be good for another round of reviews (or merge :wink:).

PR #173, opened 2017-01-13
Fix py2k bug

ocefpaf:
:weary:


ocefpaf:
The Windows failure `OSError: NetCDF: I/O failure` seems to be intermittent since we are downloading external data in that notebook. I also has to switch the `OS X` testing to the `allow_failures` section in https://github.com/ioos/notebooks_demos/pull/173/commits/71639b151a6e64141f75bca3a165e309a28ca53d of Travis-CI due to constant timeouts.

PR #174, opened 2017-01-13
Fix html table

ocefpaf:
The fancy tables do not play well with the page CSS, so we need to removed any fancy styling from the notebook.

PS: I also pulled commit from https://github.com/ioos/notebooks_demos/pull/173 to check the CIs.


ocefpaf:
Again the Windows failure seems to be an intermittent issue with the external data. Merging this to fix the page tables.

PR #180, opened 2017-01-24
env file for R

ocefpaf:
Not ready yet! Adding this jsut to start a discussion on how to proceed. (Notebook will be up soon.)

This extra env file creates the `R` environment separate from the `IOOS` environment. Ideally we should install everything in the same env to avoid convoluted logic in Travis-CI and risking having the files out of sync. Not mentioned that we won't be able to cross language barriers by accessing `Python` and `R` and vice-versa.

The reason for that is because we cannot install `R` along `conda-forge` packages right now due to some mismatch in the dependency version. In theory https://github.com/conda-forge/r-base-feedstock/pull/5 should fix that and we could try again adding all `R` packages in a single env with `Python`.


ocefpaf:
@jbosch-noaa this is how the R-notebooks looks like:

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/R-notebook/notebooks/2017-01-23-R-notebook.ipynb

There is nothing interesting there. I just played with the `r-oce` library (the only library I know in `R` :wink:)

ocefpaf:
@jbosch-noaa I added some text to the notebook and, right now, some explanation on the two env file situation. I will update it if I can get to a single env file, but this is ready for review.

PS: The r-notebook *per se* is quite weak because I am not a `R` user. I just added some instructions on how to get it working and showed some examples.

jbosch-noaa:
This notebook looks good as a simple example on how to use a Jupyter notebook with R.  I have no edits!

PR #182, opened 2017-02-07
remove anaconda-navigator from environment.yml

rsignell-usgs:
We don't need anaconda-navigator in the environment. 
 It should instead by installed only in the miniconda base environment

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,6 @@ dependencies:
   - icu=56
   - libnetcdf=4.4.0
   - python=3.5
-  - anaconda-navigator
   - bagit
   - beautifulsoup4
   - bokeh


ocefpaf:
> It should instead by installed only in the miniconda base environment

But we cannot do that with an env file. I also don't know if the navigator is pre-installed on miniconda... I guess we need to edit the instructions again...

environment.yml
```diff
@@ -6,7 +6,6 @@ dependencies:
   - icu=56
   - libnetcdf=4.4.0
   - python=3.5
-  - anaconda-navigator
```
ocefpaf: Can you remove that from the Python 2 env file too?

PR #186, opened 2017-02-15
Test Python 3.6

ocefpaf:
Upgrade the default environment to `Python 3.6`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,11 +30,6 @@ install:
   - conda env create --quiet --file "%ENV%"
   - activate "%ENV_NAME%"
   - conda install console_shortcut
-  # Fortran cells.
-  - conda install --channel conda-forge mingwpy
-  - set CFG=%CONDA_PREFIX%\Lib\distutils\distutils.cfg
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -30,11 +30,6 @@ install:
   - conda env create --quiet --file "%ENV%"
   - activate "%ENV_NAME%"
   - conda install console_shortcut
-  # Fortran cells.
```
ocefpaf: We don't have any `Fortran` notebook yet and this does not work with latest Python. So I am removing this and in the future we can experiment with `msys2` instead.

ocefpaf:
The failure on Windows seems to be a time out as sometimes passes. So merging this...

lukecampbell:
Is there such a thing as a fortran notebook?

ocefpaf:
> Is there such a thing as a fortran notebook?

You can use both the Fortran magic in cells as well a language kernel.
See this example: https://github.com/ZedThree/jupyter-fortran-kernel/blob/master/example-notebook.ipynb

lukecampbell:
whoa

PR #187, opened 2017-02-18
Single env for Python and R kernels

ocefpaf:
@rsignell-usgs and @jbosch-noaa with this change we can have a single env file and install two kernels: `R` and `Python`.

This is now possible thanks to PRs: https://github.com/conda-forge/r-base-feedstock/pull/10, https://github.com/conda-forge/staged-recipes/pull/2309, https://github.com/conda-forge/staged-recipes/pull/2456, https://github.com/conda-forge/staged-recipes/pull/2461, and https://github.com/conda-forge/staged-recipes/pull/2460.

All you need to do is to switch the `kernel` from `Python` to `R` in the top left menu.


environment-python-2.yml
```diff
@@ -70,3 +70,7 @@ dependencies:
   - thredds_crawler
   - xarray
   - xlrd
+  # R packages.
+  - r-irkernel
```
ocefpaf: Right now we have only the `R` kernel and its dependencies, and two oceanographic libraries. We should get suggestions from `R` users for more packages to add.

environment-r.yml
```diff
@@ -1,11 +0,0 @@
-name: R
-channels:
-  - defaults
-  - r
-dependencies:
-  - psutil
-  - nb_conda_kernels
-  - jupyter
-  - r-essentials
```
ocefpaf: We no longer need a separate env now that the packages are consistent with `conda-forge` versions. However, we still do not have `r-essentials`. That is a collection of the basic `R` packages an average users expects. I will work on that soon.

ocefpaf:
All green and both kernels are tested now.

```
[Running notebook]: 2017-01-18-siphon-explore-thredds.ipynb
Kernel: python
[Passed]

[Running notebook]: 2017-01-23-R-notebook.ipynb
Kernel: R
[Passed]
```

PR #188, opened 2017-02-19
remove binder and update readme

ocefpaf:
@jbosch-noaa I removed the `binder` badge b/c, unfortunately, binder became an unreliable service. (I am looking into alternatives.)

I also added more metadta, most importantly a license. Can you please check if the CC by is an OK license for `IOOS`?

You can see a rendered version at: https://github.com/ocefpaf/notebooks_demos/blob/readme/README.md


jbosch-noaa:
@ocefpaf - This is good but after checking with Derrick, he advises we should probably hold off on the license for now.  Its a sticky issue the we (IOOS) haven't nailed down a policy for yet.

ocefpaf:
Sure, I'll remove the license entry soon. Hold off the merge for now.

ocefpaf:
@jbosch-noaa I removed the license part and this should be good to go.

(The failures are related to old notebooks trying to access data that is no longer available. My testing planned for that so we could easily find "rotten" examples, but now I see it will be quite annoying to keep them updated all the time...)

Issue #172, opened 2017-01-11
How to reproduce "Boston Light Swim" for different geography


robragsdale:
I have an idea for a notebook to predict water temperatures around Hatteras Island and in Pamilico sound to help a sea turtle stranding network anticipate water temperatures dropping below 50°F that can cause a "cold-stunned event" which is when sea turtles become unable to swim and wash ashore.

ocefpaf:
@robragsdale in theory you should be able to change the bounding box and just run it. However, there are many moving parts in that notebook and I believe it won't be that simple.

We should start from the latest, not merged yet, version form #130 

Do you have the bounding box and times for this example so I can try to put an example together for you?

robragsdale:
@ocefpaf the bbox [-75.42, 35.12, -75.33, 35.48] for start: 2016-01-04 00:00:00 stop: 2016-01-07 00:00:00.  Thanks for helping with an example.

ocefpaf:
@robragsdale here is my first attempt:

http://nbviewer.jupyter.org/gist/ocefpaf/4b5e261dc0a6603ef6bd194c87d94809?flush_cache=true

Note that I did not use `pyoos` to find SOS data because I wanted to create a `100%` catalog version of this. (I could not find NDBC nor CO-OOPS buoys anyways.)

Note that there is not much to work with with those constraints.

Gridded data:
- http://oos.soest.hawaii.edu/thredds/dodsC/hioos/satellite/dhw.html
- http://thredds.aoos.org/thredds/dodsC/G1_SST_GLOBAL.nc.html
- http://oos.soest.hawaii.edu/thredds/dodsC/hioos/model/atm/ncep_global/NCEP_Global_Atmospheric_Model_best.ncd.html
- http://tds.maracoos.org/thredds/dodsC/SST-Seven-Agg.nc.html
- http://oos.soest.hawaii.edu/thredds/dodsC/pacioos/hycom/global.html

Trajectory data (glider):
- http://data.ioos.us/thredds/dodsC/deployments/drudnick/sp066-20151217T1624/sp066-20151217T1624.nc3.nc.html
- http://data.ioos.us/thredds/dodsC/deployments/drudnick/sp065-20151001T1507/sp065-20151001T1507.nc3.nc.html

rsignell-usgs:
Since this goal was to **predict** water temperatures, how about the last 5 days and next 3 days.  That should at least turn up a few of the same forecast models we got in the Boston Light Swim example

[...]

PR #189, opened 2017-02-19
Hatteras Islands notebook

ocefpaf:
This addresses @robragsdale suggestion from  #172

Right now the search is too strict and we cannot find any observations, only models.

For HTML comments/reviews please go to: https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/hatteras-islands/notebooks/2017-02-20-hatteras-islands.ipynb


ocefpaf:
Closing this for now but we can re-visit it at a later time once we figure out the data time/space range for a positive search results.

PR #191, opened 2017-03-03
Point Dockerfile at IOOS3

rsignell-usgs:
The dockerfile was still pointing at IOOS instead of IOOS3


ocefpaf:
Will this binder image work with 'everywhere'? Let's find out!

Issue #175, opened 2017-01-18
"Suggest a Demo" button - aka- how to open a GitHub issue


jbosch-noaa:
We should put a "Suggest a Demo" button on our site. Perhaps in the sidebar.  The button would link to our GitHub repo.  

We can also include a separate link on how to open a GitHub issue.  @ocefpaf mentioned that GitHub has some templates for issues and we can create an example to guide people how to do it.

jbosch-noaa:
@ocefpaf  - What do you think of this icon to put across the top corner of our pages? We can use it to link to our GitHub repo.

![suggestdemo png](https://cloud.githubusercontent.com/assets/13106516/22388218/747e5a06-e4ad-11e6-8c7c-20ccfc9b9bc2.png)

ocefpaf:
That is awesome! Thanks for putting this together @jbosch-noaa. I am looking how to best add that without adding it to the theme, because that is specific to this site.

jbosch-noaa:
@ocefpaf - here is the website I found the icon on.  That might help?

http://tom.preston-werner.com/

ocefpaf:
This is done. Closing...

PR #195, opened 2017-03-10
Suggest a notebook template

ocefpaf:
@jbosch-noaa this does not add the ribbon from #175 yet because we need to modify the theme. See https://github.com/ioos/ioos_jekyll_theme/pull/26

Right now it only adds text boilerplate, that users will see when opening an issue via the link, and the ribbon configuration. (I guess we need to warn users regarding "regular" issues.)

Here is how it looks:

ribbon

### DIFF
--- a/webpage/_config.yml
+++ b/webpage/_config.yml
@@ -121,6 +121,12 @@ author:
   xing             :
   youtube          :
 
+forkme:
+  active: true
+  title: Suggest a Demo!
+  url: https://github.com/ioos/notebooks_demos/issues/new
+  position: top-right
+
 # Reading Files
 include:
   - .htaccess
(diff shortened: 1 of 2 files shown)


webpage/_config.yml
```diff
@@ -121,6 +121,12 @@ author:
   xing             :
   youtube          :
 
+forkme:
+  active: true
+  title: Suggest a Demo!
+  url: https://github.com/ioos/notebooks_demos/issues/new
+  position: top-right
+
```
ocefpaf: This part needs https://github.com/ioos/ioos_jekyll_theme/pull/26 merged first.

ocefpaf:
> @jbosch-noaa this does not add the ribbon from #175 yet because we need to modify the theme. See ioos/ioos_jekyll_theme#26

We need to merge ioos/ioos_jekyll_theme#26 too and then rebuild the pages otherwise it won't work.

jbosch-noaa:
sorry!  I miss read the note.

ocefpaf:
No problem. It won't break the page. It just won't add the ribbon.

PR #201, opened 2017-03-29
xtractoR

ocefpaf:
This is a `R` notebook that reproduces some of the `xtractomatic` examples.

`hypothes.is` link: https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/xtractoR/notebooks/2017-08-01-xtractoR.ipynb


ocefpaf:
Phew! This was a difficult one to solve for the dependencies!!

https://github.com/conda-forge/r-ncdf4-feedstock/pull/4
https://github.com/conda-forge/r-xtractomatic-feedstock/pull/1
https://github.com/conda-forge/staged-recipes/pull/3392
https://github.com/conda-forge/r-maps-feedstock/pull/2
https://github.com/conda-forge/r-mapdata-feedstock/pull/1
https://github.com/conda-forge/r-rerddap-feedstock/pull/1

But is is ready for review now :smile:

jbosch-noaa:
@ocefpaf 
Ok I made some comments.  Let me know what you think.  :)

ocefpaf:
> Ok I made some comments.  Let me know what you think.  :)

I implemented most of the review request and we discussed a few others in the `hypothes.is`, but here is what I was unable to comply 100%:

- the colormap for the `scale_colour_gradient` accepts only a low/high value to create the gradient. So I cannot use rainbow or something similar. I did reverse the default though to make it look better.
- I did change the text for the `xlen`/`ylen` explanation b/c it will be degrees in this dataset, but I do believe the "correct" way would be to write data units instead of always assuming degrees. (I checked the `xtractomatic` code and I don't see any conversion.)

jbosch-noaa:
OK .. the edits will work.  Thanks for attempting to fix everything.  Great work getting this notebook to work with r !

PR #203, opened 2017-03-29
Add a note to explain `ioos_tools`.

### LINKED ISSUE #192: CSW demo can't be downloaded and run
The [CSW example notebook](http://ioos.github.io/notebooks_demos/notebooks/2016-12-19-exploring_csw/) has a "download" at the bottom, but if you do that you can't run it, even if you have the IOOS environment, because it depends on 
```
ioos_tools = os.path.join(os.path.pardir)
sys.path.append(ioos_tools)
```

ocefpaf:
Closes #192 

@rsignell-usgs can you take a look at the text I added to explain the `ioos_tools` dependency?

```
This notebook uses a custom `ioos_tools` module that needs to be added to the
path separately. We recommend cloning the
[repository](https://github.com/ioos/notebooks_demos)
on GitHub which already includes the most update version of `ioos_tools`.
```

Any suggestions?

### DIFF
--- a/notebooks/2016-10-12-fetching_data.ipynb
+++ b/notebooks/2016-10-12-fetching_data.ipynb
@@ -3,7 +3,9 @@
   {
    "cell_type": "markdown",
    "metadata": {
-    "collapsed": false
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
    },
    "source": [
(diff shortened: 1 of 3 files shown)


ocefpaf:
Merging this, if anyone has any suggestions we can send another PR.

PR #204, opened 2017-03-29
Matlab/Octave demo

ocefpaf:
This is a very simple `Matlab`/`Octave` notebook demostration.

I did want to make this a copy of https://github.com/ioos/notebooks_demos/pull/201 using `Matlab` instead of `R`. However, I am limited to running only code that is compatible with `Octave`, because `Matlab` is non-free. Unfortunately `xtractomatic` is not compatible with `octave` (nor even old versions of Matlab).

Note that I don't use Matlab for more than 10 years now and I am unaware of interesting toolboxes that users might want to see. Please ping me with suggestions!

Here is the `hypothes.is` link: https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/matlab_octave/notebooks/2017-03-30-octave_notebook_example.ipynb?flush_cache=true

### DIFF
--- a/notebooks/2017-03-30-octave_notebook_example.ipynb
+++ b/notebooks/2017-03-30-octave_notebook_example.ipynb
@@ -1,0 +1,343 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "A common misconception is that `Jupyter` notebooks are for Python only.\n",
+    "We already showed how to use the\n",
+    "[`R` language](http://ioos.github.io/notebooks_demos/notebooks/2017-01-23-R-notebook/),\n",
+    "and now we are going to create a Matlab/Octave toolboxes notebook.\n",
+    "\n",
+    "The IOOS conda\n",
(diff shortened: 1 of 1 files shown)


jbosch-noaa:
Thanks for taking a stab at this @ocefpaf .  Actually, I think @mbiddle-ndoc might have some good input here.  
Matt - What do you think?  Do you have a good IOOS data / oceanographic matlab example?

ocefpaf:
@jbosch-noaa I also fixed a number of CI failures in this PR. Let me know if you are OK with this Matlab/Octave notebook example. If not I will pull the commits from the CI fixes into another PR so safe them.

jbosch-noaa:
I looked over this notebook and I think it looks good.  This is an appropriate first notebook for Matlab.  It sets the stage for more Matlab notebooks in the future. :)

jbosch-noaa:
Oh no!  I just saw that this notebook doesn't have a title on the thumbnail!!  @ocefpaf - would you please fix that. Perhaps title it "Simple Matlab/Octave Demo"

ocefpaf:
I'll fix this later today. (Got stuck in Atlanta last night due to weather
and I did not make it to Boston yet.)

PR #209, opened 2017-04-04
Fix MBON video URL

### LINKED ISSUE #208: MBON tutorial link update
@ocefpaf  - The MBON portal demo link has just been updated by @bbest.  Would you please update it on our video tutorial part of the website?  Here is the new link:

    youtube.com/watch?v=ZITqDRa6u9c

ocefpaf:
Closes #208 

@jbosch-noaa we can use iframes to allow people to click and watch the video without leaving the page.
I will try a prototype of that and let you know.

### DIFF
--- a/webpage/video_tutorials.md
+++ b/webpage/video_tutorials.md
@@ -9,7 +9,7 @@ layout: single
 
 - [COMT Tutorial](https://www.youtube.com/watch?v=Dqc1C1HeemQ)
 - [IOOS EDS Demo](https://nccospublicstor.blob.core.windows.net/ioos/ioos_demo_1280.mp4)
-- [MBON Portal Tutorial](https://www.youtube.com/watch?v=StpS6RPECzM)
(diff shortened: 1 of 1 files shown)


jbosch-noaa:
We might not want them to stay on the page because the iframe would be too small to see the video. 

It would be good to have the link open a new tab or browser window though.  In fact we should do that for the notebooks also, this way people wouldn't leave our Demo Center page.

ocefpaf:
> We might not want them to stay on the page because the iframe would be too small to see the video.

We can control the size and people can always toggle "fullscreen", it would not be different from watching it on youttube directly. The only downside I foresee is that people would need to scroll further the page to see everything instead of seeing the list of links all together.

>  In fact we should do that for the notebooks also, this way people wouldn't leave our Demo Center page.

But we do render the notebooks as a page. Do you mean the link to download/view the raw notebook?

ocefpaf:
@jbosch-noaa merging this to get the fix in the repo and we can keep discussing the iframe vs links in a new issue.

PR #212, opened 2017-04-05
Video IFrames

ocefpaf:
@jbosch-noaa this is how the page will look like with the video IFrame.

site

I am on the fence b/c I like the simple listing we have now but I understand that some people would prefer to watch the videos directly instead of being re-directed to youtube. What do you think? Close it or merge it?

### DIFF
--- a/webpage/video_tutorials.md
+++ b/webpage/video_tutorials.md
@@ -8,15 +8,33 @@ layout: single
 ![IOOS-logo_little]({{ site.url }}/{{ site.baseurl }}/assets/IOOS-logo_little.png) IOOS
 
 - [COMT Tutorial](https://www.youtube.com/watch?v=Dqc1C1HeemQ)
+
+<iframe id="video" width="420" height="315" src="//www.youtube.com/embed/Dqc1C1HeemQ?rel=0" frameborder="0" allowfullscreen></iframe>
+
 - [IOOS EDS Demo](https://nccospublicstor.blob.core.windows.net/ioos/ioos_demo_1280.mp4)
+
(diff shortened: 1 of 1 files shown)


jbosch-noaa:
I am also kind of torn.  I also like just the list and fear as we create/collect more demo videos the site will get too crowded.  Let me think on it ...

jbosch-noaa:
Ok @ocefpaf , I've thought about it an think we should try it.  Have you seen the way they do videos on this site?

http://oceanslive.org/video-gallery/itemlist/category/57-capitol-hill-ocean-week-2016

Can we do the videos in 2 columns?  I can write short blurbs about the videos as well. ...and when I say short, I mean 1-2 sentences at most. 

@mwengren suggests 2 columns can be done without changing the template by using inline css code with 2 div tags (one for each column).

mwengren:
I think you could accomplish something like that with 

<div style="float: left">

</div>
<div style="float: right">

</div>

There may need to be more CSS tricks involved to get it to work (fixed width etc), and it may cause issues with the responsive design of the template, not sure.

ocefpaf:
Sure. I will try that next week.

ocefpaf:
@jbosch-noaa I tried a few options here and none looked good in some the videos became too small because of the sidebar while in others they jumped from the side to the bottom of the neighbor video.

Also, the videos are not all from youtube and not all with the same aspect ratio, making this even harder to adjust in a generic way for all videos.

I believe that we either keep the original list from of merge this PR as is.

jbosch-noaa:
Ok.  Lets keep it as is for now.  Thanks for trying!

PR #221, opened 2017-04-21
Reduce the CI run time

ocefpaf:
The CIs are taking too long and as we accumulate more notebooks it will only get longer.
This PR tries to alleviate that by reducing the AppVeyor build matrix and caching the environment on Travis-CI.

Another idea is to move old notebooks to a `allow_failure` section with `fast_finish`. That way we would cover only the new submitted notebook in every new PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,74 +2,105 @@ language: generic
 
 sudo: false
 
+cache:
+  directories:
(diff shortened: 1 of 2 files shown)


ocefpaf:
:tada: first time both CIs are green in a long time :tada:

PR #222, opened 2017-04-21
eliminating need for ioos_tools in csw notebook

rsignell-usgs:

PR #223, opened 2017-04-21
Re-based alternative to #222

ocefpaf:
Same as #222 but rebased and without the execution time text in the cells.

@rsignell-usgs can you take a look at https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/222_alt/notebooks/2016-12-19-exploring_csw.ipynb and check if I got all the text you modified?


ocefpaf:
Merging this to get the changes into the page but we can edit it later if that is not OK yet.

PR #225, opened 2017-05-04
Update gamone URL to https

ocefpaf:
Ping @rsignell-usgs


rsignell-usgs:
Did the http URLs no longer work?  They should have automatically switched to https, no?

ocefpaf:
In #224 I added a fail/pass to the catalogs that do not respond due to the maintenance today. That is unrelated to the re-direction.

In this PR I decided to explicitly add the `https` instead. But yes, the `http` are probably being correctly re-directed to `https` now.

rsignell-usgs:
okay, cool.

PR #229, opened 2017-06-06
Fix bad dimension promotion with safer logic

ocefpaf:
@rsignell-usgs this solves the second issue and the Boston Light Swim should be running again.


ocefpaf:
`[PASSED]: 2016-12-22-boston_light_swim.ipynb` :tada: 

(but `2016-12-19-exploring_csw.ipynb` is failing now :unamused:)

PR #234, opened 2017-06-20
Add utide to the env file

ocefpaf:
No real change to the page or the notebooks, just making the env more complete for oceanographers.


ocefpaf:
The failure is from an old notebook and unrelated to this PR.

PR #235, opened 2017-06-20
Binder

ocefpaf:
@jbosch-noaa and @rsignell-usgs thanks to the awesome work of @choldgraf, @yuvipanda, and many others, we can restore the binder badge with an extra feature! Now we have a link to a live notebook in the footer of every page. Users should be able to just lick there and run.

I replaced the static `nbviewer` link with the live binder one to avoid polluting the footer.

PS: this version of binder is still beta, but quite stable and more reliable than the old binder at this moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,10 @@ See the rendered version at https://ioos.github.io/notebooks_demos
 
 To suggest a notebook or ask questions please open an issue at: https://github.com/ioos/notebooks_demos/issues
 
+## Binder
+
+[![Binder](http://mybinder.org/badge.svg)](https://beta.mybinder.org/v2/gh/ioos/notebooks_demos/master)
+
 ## Citation
 
 [![DOI](https://zenodo.org/badge/58492405.svg)](https://zenodo.org/badge/latestdoi/58492405)
--- a/jupyter-jekyll.tpl
+++ b/jupyter-jekyll.tpl
@@ -88,5 +88,5 @@ unknown type  {{ cell.type }}
 <br>
 Right click and choose Save link as... to
 [download](https://raw.githubusercontent.com/ioos/notebooks_demos/master/notebooks/{{resources['metadata']['name']}}.ipynb)
-this notebook, or see a static view [here](http://nbviewer.ipython.org/urls/raw.githubusercontent.com/ioos/notebooks_demos/master/notebooks/{{resources['metadata']['name']}}.ipynb).
+this notebook, or click [here](https://beta.mybinder.org/v2/gh/ioos/notebooks_demos/master?filepath=notebooks/{{resources['metadata']['name']}}.ipynb) to run a live instance of this notebook.
 {%- endblock footer -%}


choldgraf:
woooah - this is a cool idea to integrate w/ jekyll like this.

also of interest perhaps is a PR I've got open in sphinx-gallery right now to auto-generate binder badges for python examples: https://github.com/sphinx-gallery/sphinx-gallery/pull/244

ocefpaf:
> also of interest perhaps is a PR I've got open in sphinx-gallery right now to auto-generate binder badges for python examples: sphinx-gallery/sphinx-gallery#244

I stole the how to create the links form that PR :smile: Thanks!

I would love to use sphinx-gallery at some point, but we are limited to `ipynb` files here because we need to embed lots of markdown and use different languages. As far as I know sphinx-gallery works only wth `.py ` and ReST.

jbosch-noaa:
Ok cool! Is this ready to be merged?

ocefpaf:
> Ok cool! Is this ready to be merged?

Should be. No need to wait for the CIs here.

choldgraf:
@ocefpaf yep - and my guess is that sphinx-gallery won't change that any time soon. Do you auto-convert your ipynb files to markdown in building your jekyll site?

ocefpaf:
>  Do you auto-convert your ipynb files to markdown in building your jekyll site?

Yep. That is the gist of it but the process is more ["complex"](https://github.com/ioos/notebooks_demos/blob/master/.travis.yml#L96-L101) than I would like, with too many moving parts. It involves a [custom jekyll](https://github.com/ioos/ioos_jekyll_theme) theme, a modified `nbconverter` template [jupyter-jekyll.tpl](https://github.com/ioos/notebooks_demos/blob/master/jupyter-jekyll.tpl), and a [python script](https://github.com/ioos/notebooks_demos/blob/master/webpage/make_index.py) to create the index. Sphinx-gallery would greatly simplify this process, but maybe I should be focusing on create a generic jekyll theme instead.

choldgraf:
I'm helping an instructor at CU boulder accomplish basically the same thing: http://earthdatascience.org/

Right now I have a custom python script that does the ipynb -> md conversion before building the site. I decided not to mess with nbconvert templates because it wasn't the easiest to wrap my head around. Would love to chat about what has worked for you / what hasn't, and maybe put some time into a jekyll ipynb template

ocefpaf:
> I'm helping an instructor at CU boulder accomplish basically the same thing: http://earthdatascience.org/

Cool site! I tried to get a similar functionality like that side bar but failed.

> Right now I have a custom python script that does the ipynb -> md conversion before building the site. I decided not to mess with nbconvert templates because it wasn't the easiest to wrap my head around. Would love to chat about what has worked for you / what hasn't, and maybe put some time into a jekyll ipynb template

Sure. Feel free to ping me anytime. I bet I will learn more from you than the other way around though :wink: 

PS: the best way to reach me is via e-mail (ocefpaf at gmail) or `gitter` (`ocefpaf`).

choldgraf:
Cool! I will try to follow up once we put together the first iteration of these materials. I think there would be value in creating something that is more general so that people can reuse and build on top of it, but I'm not a super expert with Jekyll so lots of learning to do still

PR #244, opened 2017-07-03
folium thumbnails

ocefpaf:
This PR uses the dev version of `folium` and its PNG `__repr__` to create a thumbnail for the gallery. That way we will use the IOOS logo only on text only notebooks like the `compliance-checker` one.

screenshot from 2017-07-03 17-20-33

I would like to wait the `folium` release to merge this to avoid failures.


jbosch-noaa:
This looks great.

PR #249, opened 2017-07-25
add hardcoded HFRadar endpoint

ocefpaf:
@jbosch-noaa there is no text yet but I wanted your :+1: before I move ahead and start creating a story about this notebook. Do you think general approach is OK?

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/HFRadar/notebooks/2017-07-25-HFRadar_currents.ipynb


jbosch-noaa:
We should probably reference something from NOAA rather than Wikipedia or Treehugger:
http://response.restoration.noaa.gov/oil-and-chemical-spills/significant-incidents/deepwater-horizon-oil-spill

Another valuable link: 
https://restoreactscienceprogram.noaa.gov/

I am also looking to elevate the HFR metadata issue with SCCOOS next week at their DMAC webinar meeting with our office.  We should really get that fixed.

ocefpaf:
> We should probably reference something from NOAA rather than Wikipedia or Treehugger

Thanks for the resources! (But ignore that text. It is a copy-n-past from the http://cordc.ucsd.edu/projects/mapping/maps/fullpage.php popup.)

jbosch-noaa:
@ocefpaf - I ran through the notebook and smoothed out the comments.  It looks good! Once you update the comments feel free to merge it.

ocefpaf:
@jbosch-noaa read for another round of reviews. Thanks!

jbosch-noaa:
@ocefpaf  - Not sure how I missed this before we published it but I looking at this notebook on the website I found a few things that we should fix.

1. The Deepwater Horizon Oil Spill happened in 2010 but we are showing data from 2017.  Wouldn't it be better if we had data from 2010?

2. At the bottom of the notebook, we have a plot of the east coast of the US.  Why? There is no relevance to the east coast.  We should have one of the Gulf of Mexico.

ocefpaf:
> The Deepwater Horizon Oil Spill happened in 2010 but we are showing data from 2017.  Wouldn't it be better if we had data from 2010?

We can do that if you think it is best. My idea was to illustrate the need for near-real time currents with the DWH accident but to actually show how to get near-real time current and not past that.

Let me know if you want to make the change and I'll adapt the text and re-run it.

> At the bottom of the notebook, we have a plot of the east coast of the US. Why? There is no relevance to the east coast. We should have one of the Gulf of Mexico.

Do you want the whole Gulf as the first panel and keeping the second or show only the second panel?

jbosch-noaa:
I think we should do an example of each.  This notebook we can highlight how to get historical data from 2010. To answer the second question above.  Lets keep the Gulf panel only. We can write a second notebook, perhaps dealing with hurricane Irma (in the Caribbean now and threatening Florida), where we highlight getting more real time or near real time data.  We can do a before and , maybe during, and after notebook for Irma.  Possibly we can show more than just HFR currents but salinity or buoy data as well.  Irma is forecasted to hit Florida this coming Saturday.

ocefpaf:
Perfect. Addressing it now. Thanks!

Regarding future work. I'll try my best to send new code, but I am limited these days. (I can only work from bed, my sciatica decided to attack me hard this week b/c the genius here moved a table by himself.)

jbosch-noaa:
Oh no!!!  Take care of yourself first and get to the new notebook when you are stronger.

ocefpaf:
>This notebook we can highlight how to get historical data from 2010. 

The time coverage for the endpoint we are using goes back only up to 2012 :unamused: 

```
TimeCoverage:

Start: 2012-01-01T00:00:00Z
End: 2017-09-05T17:02:22.604Z
```

See http://hfrnet.ucsd.edu/thredds/HFRADAR_USEGC_hourly_RTV.html?dataset=HFRNet/USEGC/6km/hourly/RTV

I check `sdf.ndbc` as an alternative but that goes only back a month. See https://gist.github.com/ocefpaf/25b6342c01d84845634b439589b97c49

Do we have an archive that goes back to 2010? (Ping @rsignell-usgs here.)

PR #257, opened 2017-09-10
hurricane irma

ocefpaf:
@jbosch-noaa this is an improved version of the notebook I sent to you via e-mail.

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/irma/notebooks/2017-09-09-hurricane_irma.ipynb

Let me know what you think, we can run it again tomorrow to update with the latest prediction from NHC.


jbosch-noaa:
@ocefpaf  Looks great!  Thanks for putting this together.  I made a few comments and clarifications.

ocefpaf:
> @ocefpaf  Looks great!  Thanks for putting this together.  I made a few comments and clarifications.

All addressed in the latest commit. Let me know if you want a second review or if this is OK to go.

PS0: should we do something similar for José?
PS1: I'm trying to get significant waive height in the plots too, but probably not for this version.

jbosch-noaa:
This one should be a go.

We could try one for Jose but only if it threatens the US. Right now it might not be an issue.

PR #260, opened 2017-09-13
hurricane code based search

ocefpaf:
@jbosch-noaa I got a suggestion from @rsignell-usgs to use only the Hurricane code info to download the GIS data, and find the dates and the bounding box to fetch the observations. That way the user has to tweak only the hurricane code to run this for a different scenario.

I also polished it a little bit and removing the "slim" version of the map by showing the full HTML version served in the gh-pages.


jbosch-noaa:
Ok.  That sounds like a great suggestion. 👍

jbosch-noaa:
I like how you annotated the eye or the hurricane on the plot :)

ocefpaf:
The Iframe approach failed to render :unamused: 

I'll try to fix that in a moment...

jbosch-noaa:
Also, I thought you were getting rid of the slim version but I noted the comment noting the slim version is still in the notebook above ln [16] :

ocefpaf:
Thanks, I'll fix that one too.

rsignell-usgs:
Perhaps we can ask Brian tomorrow if  JuptyerLab could make this type of notebook a live web app.

ocefpaf:
`ipyleaflet` can probably do most of that but I have to confess that I never tried it.

ocefpaf:
> Also, I thought you were getting rid of the slim version but I noted the comment noting the slim version is still in the notebook above ln [16] :

@jbosch-noaa I cannot find it, maybe you were looking at the older version before it refreshed. Here is the current `cell [16]`:

```python
In [16]:
index = pd.date_range(
    start=start.replace(tzinfo=None),
    end=end.replace(tzinfo=None),
    freq='15min'
)

# Re-index and rename series.
ssh_observations = []
for series in ssh_obs:
    _metadata = series._metadata
    obs = series.reindex(index=index, limit=1, method='nearest')
    obs._metadata = _metadata
    obs.name = _metadata['station_name']
    ssh_observations.append(obs)

winds_observations = []
for series in win_obs:
    _metadata = series._metadata
    obs = series.reindex(index=index, limit=1, method='nearest')
    obs._metadata = _metadata
    obs.name = _metadata['station_name']
    winds_observations.append(obs)
```

jbosch-noaa:
Ok.  I guess it was an older version.  It looks good now! :)

ocefpaf:
I'm fixing other `IFrame`s that are also broken in my next PR. I found 3 other posts with the same issue :sob:

PR #265, opened 2017-09-14
@rsignell-usgs suggestions

ocefpaf:
@rsignell-usgs I implemented two of your requests:

- no more served external HTML, I found a solution to bundle the generated HTML and by pass the `Jupyter` notebook `__repr__` limitation;
- added a hover `tooltip` to the `bokeh` plots.

The `tooltip` shows the variable name and data y-value. I could not get bokeh date formatting to work to show the x-value.

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/rsignell-usgs_suggestions/notebooks/2017-09-09-hurricane_irma.ipynb

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,4 +30,5 @@ install:
   - conda list
 
 test_script:
-  - cmd: cd tests && python test_notebooks.py
+  - cmd: cd tests
+  - cmd: python test_notebooks.py
--- a/webpage/other_resources.md
+++ b/webpage/other_resources.md
@@ -15,7 +15,7 @@ you can install it by typing `conda install anaconda` using miniconda.
 ## Install
 
 Download and install the appropriate Miniconda installer from [http://conda.pydata.org/miniconda.html](http://conda.pydata.org/miniconda.html).
-With Anaconda you can create environments that use any Python version (e.g. Python 2.7 or Python 3.x),
+With Anaconda you can create environments that use any Python version (e.g. Python 2.7 or Python 3.6),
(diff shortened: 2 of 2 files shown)


rsignell-usgs:
I would still prefer to remove the hardwired aspects of cell [16], just picking a random station.   
Or just delete it.   
Or move it to the end.

ocefpaf:
> I would still prefer to remove the hardwired aspects of cell [16], just picking a random station.
Or just delete it.
Or move it to the end.

The original idea was a notebook focused on Irma that is why I added that close inspection to one of the stations. I don't really have a strong opinion.

What do you think @jbosch-noaa?

jbosch-noaa:
I think we should leave it.  I am already advertising this notebook as showcasing Irma data.  If we can add more Irma data the better.

rsignell-usgs:
That's fine.  I can always delete it when I'm showing it.    My pitch will be that it works for Irma, but the real power is that you can change the number and it will work for any storm, which is very cool (we are now interested in Jose!)

ocefpaf:
> That's fine. I can always delete it when I'm showing it.

I deleted it from my own [1] non-blog version.
B/c  this one here is focused on Irma I guess it is better to keep it.

> My pitch will be that it works for Irma, but the real power is that you can change the number and it will work for any storm, which is very cool (we are now interested in Jose!)

BTW, I just got a message from a friend in Cuba who is running it for `al142017`.
I guess that José is New England's problem now.

[1] https://github.com/ocefpaf/python_hurricane_gis_map

PR #268, opened 2017-10-03
r-erddap notebook

### LINKED ISSUE #165: Python & R Notebook
Write a notebook showing how to access data with Python and then use R to analyze that data.

ocefpaf:
~~# Not ready for merge yet!~~

@jbosch-noaa do you believe a notebook like this is worth adding to the data demo center?

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/rerddap/notebooks/2017-10-09-rerddap.ipynb

Ready for review!

Closes  #165


jbosch-noaa:
This isn't bad but I don't think we should recreate Roy's notebook.  If anything, I think we should put a link to Roy's notebook on our site, in the side bar. 

If an R notebook accessing ERDDAP is our goal, perhaps we should find a different ERDDAP server to use and create a different science story.  I would still like to see a biological example.  Check out the[ GCOOS ERDDAP server ](http://gcoos4.tamu.edu:8080/erddap/info/index.html?page=1&itemsPerPage=1000) and see if you have any ideas. There is Coral Reef data there ... maybe something with that dataset?

ocefpaf:
> This isn't bad but I don't think we should recreate Roy's notebook. If anything, I think we should put a link to Roy's notebook on our site, in the side bar.

Makes sense. I wanted to give the R-notebook the same look and feel of the demo page, but that is hard. So a link is probably better.

> If an R notebook accessing ERDDAP is our goal, perhaps we should find a different ERDDAP server to use and create a different science story. I would still like to see a biological example. Check out the GCOOS ERDDAP server and see if you have any ideas.

Awesome! I'll check it today.

> There is Coral Reef data there ... maybe something with that dataset?

Will do. I'll ping you once I have something,

ocefpaf:
@jbosch-noaa I updated the notebook to use the *Coral Reef Evaluation and Monitoring Project (CREMP) V2.0 - Florida Keys*. In the server there are datasets from 1996 to 2014 but they look like the same data. Maybe they are just an updated dataset? Not sure :confused: 

The notebook uses `r-erddap` to get the dataset and python to plot it so we can address #165 too. I'll add some text if you think this is right path.

ocefpaf:
@jbosch-noaa I acted on most of your review comments. Here is a summary:

- I copied the comment on what is `quantificationValue` and added a link to the `tabledap` `info` page b/c I could not find how to obtain the variable metadata with `rerddap` by reading the docs;
- according to your explanation I should keep the `zero` values and filter only the `-999`;
- by keeping the zeros the averages changed and the largest biomass is around `Orbicella` and not `Dendrogyra`;
- the barplot in `cell [8]` has no meaning with the zeros b/c all habitats will sum up to `0.02`;
- you made this comment on the `viridis_r` line:

      Is there are reason why you chose this critter?
      You started looking at Acropora cervicornis.
      We should pick a species and stick with it.

I did not choose `Acropora cervicornis` in the notebook, everything so far was done for all species of Coral. Also, `viridis_r` is the [colormap](https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html), not a critter :wink:

You also commented:

     Perhaps we make 3 plots of the top 3 most abundant species and compare where they occur?

I can do that. Let me adjust the notebook a little bit.

ocefpaf:
@jbosch-noaa as we discussed in the `hypothes.is` review notebook I added:

- a map with the 3 genus where the higher biomass occurs;
- restored the zero values filtering and added a note on the "biased" averaging;
- restored the count per habitat.

I don't see anything interesting regarding the biology to add, but that does not mean there is nothing there. I am just not qualified to discuss Coral Reef biomass :grimacing:

jbosch-noaa:
I took a quick look and made one suggestion.  I will do a better read next week because I'm kinda busy today. :)

ocefpaf:
> I took a quick look and made one suggestion. I will do a better read next week because I'm kinda busy today. :)

Thanks for the review. No rush, I'll get back to the HF-Radar notebook and finish the work on the Python ERDDAP wrapper.

jbosch-noaa:
@ocefpaf - I made a bunch of comments on the notebook back on Oct. 13th after all.  I see you worked on it a bit yesterday but I don't see your responses to my comments.

ocefpaf:
@jbosch-noaa my last commit should address you last review this is good for another round.

I did split the last map into 3 as you request and I believe it is more interesting this way, but I am no biologist :grimacing:

PR #270, opened 2017-10-04
Add sensorml2iso module

mwengren:
Can we add this module into the default IOOS environment?

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -60,6 +60,7 @@ dependencies:
   - seaborn
   - seawater
   - selenium
+  - sensorml2iso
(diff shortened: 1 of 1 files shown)


ocefpaf:
> Can we add this module into the default IOOS environment?

Sure. We don't have any notebook that uses that but we should probably add one.
Do you have a nice example?

mwengren:
I don't have a notebook that uses it, but it is used by some RAs and we had a request yesterday to include it in the IOOS default env by one, so it would be available for their general Python use. Thanks!

PR #275, opened 2017-11-27
add bagit text

ocefpaf:
@jbosch-noaa this is @kwilcox's notebook on bagit with some text to help the reader understands what is going on in the code cells.

Here is the `hypothes.is` link for review: https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/bagit_text/notebooks/2017-11-01-Creating-Archives-Using-Bagit.ipynb

### DIFF
--- a/notebooks/2017-11-01-Creating-Archives-Using-Bagit.ipynb
+++ b/notebooks/2017-11-01-Creating-Archives-Using-Bagit.ipynb
@@ -7,7 +7,8 @@
     "# Using [`BagIt`](http://libraryofcongress.github.io/bagit-python/) to tag oceanographic data\n",
     "\n",
     "\n",
-    "BagIt is a packaging format that supports storage of arbitrary digital content. The \"bag\" consists of arbitrary content and \"tags,\" the metadata files.\n",
+    "`BagIt` is a packaging format that supports storage of arbitrary digital content. The \"bag\" consists of arbitrary content and \"tags,\" the metadata files. `BagIt` packages can be used to facilitate data sharing with federal archive centers - thus ensuring digital preservation of oceanographic datasets within IOOS and its regional associations. NOAA NCEI supports reading from a Web Accessible Folder (WAF) containing bagit archives. For an example please see: http://ncei.axiomdatascience.com/cencoos/\n",
+    "\n",
     "\n",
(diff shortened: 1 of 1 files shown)


jbosch-noaa:
This notebook looks great!  Thanks to both of you (@kwilcox, @ocefpaf ) for putting it together.  After reading the notebook I finally understand what Bagit does.  :)

ocefpaf:
There are some missing details about the meaning of the hashes, etc. But I did not want to get into that b/c I don't think most of the user would be interested.

jbosch-noaa:
That's fine about the hashes. I just think one statement at the top regarding what folks would use BagIt for in the IOOS context would be helpful.

ocefpaf:
> I just think one statement at the top regarding what folks would use BagIt for in the IOOS context would be helpful.

I actually never used it and I am not sure how `BagIt` is used in the IOOS context... But how about:

     `BagIt` is used to ensure digital preservation of oceanographic datasets within IOOS and
     its regional associations.

jbosch-noaa:
That sounds good.  Just add that and then we can merge it.

kwilcox:
`BagIt` packages can be used to facilitate data sharing with federal archive centers - thus ensuring digital preservation of oceanographic datasets within IOOS and its regional associations. NOAA NCEI supports reading from a Web Accessible Folder (WAF) containing bagit archives. For an example please see: http://ncei.axiomdatascience.com/cencoos/

ocefpaf:
I think @kwilcox's version is better :smile:

jbosch-noaa:
I agree.  Thanks @kwilcox !

ocefpaf:
> I agree. Thanks @kwilcox !

Added that. This is good to go @jbosch-noaa.

ocefpaf:
Fixed that in #278 but I I'll try to fix the gallery later so we can use links in the titles without that problem.

jbosch-noaa:
@ocefpaf -I Got this feedback from @emiliom off the ioos_tech email list and I think we should make these edits:

Bob's clarification about BagIt is an important one. I'd like to suggest a couple of small tweaks to the notebook preamble text to avoid confusion:

- Change the "BagIt" link at the start of the paragraph to point to https://en.wikipedia.org/wiki/BagIt instead of the python bagit package
- In the second paragraph, make the phrase "python interface" into a link that points to the python bagit package, http://libraryofcongress.github.io/bagit-python

Cheers,
-Emilio

Issue #241, opened 2017-07-02
Create a Wave Height Skill Assessment notebook


ocefpaf:
xref.: https://github.com/ioos/system-test/issues/209

PR #282, opened 2017-12-20
wave height assessment

ocefpaf:
~~# Not ready for review yet!~~

Closes https://github.com/ioos/notebooks_demos/issues/241

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/waves/notebooks/2018-03-30-wave_height_assessment.ipynb


ocefpaf:
@jbosch-noaa this is ready for review. Note that I did not write an explanation for each cell. The reason is b/c this notebook uses virtually the same workflow as the Boston Light Swim and the Sea Surface Height, because of that I'd rather refer the read to those rather than repeating the text here.

This notebook complements https://github.com/ioos/notebooks_demos/pull/303 and show an interesting increase in the significant waves height around March 3rd (I believe that is when Nor’easter hit the east coast in that area).

Ping @rsignell-usgs

jbosch-noaa:
Excellent!  I have no comments.  It all looks great. : )

PR #286, opened 2018-01-29
no change just to test the env creation

ocefpaf:
# Don't merge this!

It is only a test to trigger the env creation in the CIs.

@rsignell-usgs reported that someone is getting:

```
UnsatisfiableError: The following specifications were found to be in conflict:
  - iris >=1.12
  - r-mapdata
  - r-rerddap
  - rasterio
```

I could not isolate the cause on Linux b/c everything works there and I don't know if the failure happens on OS X or Windows, so let's use the CIs to check that.


ocefpaf:
Aha. Thanks to Travis-CI identified it to be OS X. Fixing it...

rsignell-usgs:
@ocefpaf , yup, it was Mac!

ocefpaf:
Alright, all the packages are available again on all platforms. Closing this.

PR #293, opened 2018-02-23
add pocean notebook demo

ocefpaf:
@jbosch-noaa this one is ready for review, here is the hypothes.is link:

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/pocean/notebooks/2018-02-27-pocean-timeSeries-demo.ipynb

@kwilcox and @rsignell-usgs I know you both are extremely busy but  I'd love your input on this notebook too if you cam spare the time :grimacing:

During the OBIS workshop and the OSM18 `pocean` was mentioned a few times and the number one complaint was docs and example. Hopefully, with this notebook, we can start to address that.


rsignell-usgs:
@ocefpaf, this is great!   I just left a few minor suggestions at:
https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/pocean/notebooks/2018-02-27-pocean-timeSeries-demo.ipynb

@kwilcox, is the "profile" number necessary?  
Seems inappropriate with a time series at a fixed depth, like this one.

ocefpaf:
> @kwilcox, is the "profile" number necessary?
Seems inappropriate with a time series at a fixed depth, like this one.

That is probably copy-n-pasta, I based this notebook [on this one](http://nbviewer.ipython.org/urls/raw.github.com/pyoceans/pocean-core/master/docs/notebooks/adcp.ipynb).

jbosch-noaa:
@ocefpaf - I made one minor comment in the beginning to offer a context as to why we are publishing this notebook.  Other than that, I agree with @rsignell-usgs  that this is good to go. :)

kwilcox:
This is great, thanks @ocefpaf. `pocean` needs some documentation outside of the API docs...

> [<class 'netCDF4._netCDF4.Variable'>
 int64 z(station)
     _FillValue: -9999
     axis: Z
     units: m
     standard_name: depth
     positive: down
 unlimited dimensions: 
 current shape = (1,)
 filling on]

pocean is outputting an `int64` for the depth variable which is going to break lots of things. I'm not handing this automatically because someone might actually need an `int64`. Can you call `df = pocean.utils.downcast_dataframe(df)` before creating the netCDF?


> Different from the standard netCDF4-python object, pocean's DSGs split the variables in data_vars, ancillary_vars, and the DSG axes.

You also have access to to the regular `.variables` so its more "In addition to" rather than "Different from"

> 't': [start + timedelta(days=n) for n in range(len(x))],
    'x': -48.6256,
    'y': -27.5717,
    'z': 10,

Might be worth mentioning that these need to be mapped to the axes of the DSG file you want to use. These just happened to be [the defaults](https://github.com/pyoceans/pocean-core/blob/master/pocean/utils.py#L50-L59). If you use other column names you need to supply the `axes` attribute to the `from_dataframe` function that maps the axis type to the column. Then again, this probably belongs in the `pocean` documentation as well.

```
df = pd.DataFrame({
    'time': [start + timedelta(days=n) for n in range(len(x))],
    'longitude': -48.6256,
    'latitude': -27.5717,
    'depth': 10,
    'u': np.sin(x),
    'v': np.cos(x),
    'station': 'fake buoy',
})
dsg = OrthogonalMultidimensionalTimeseries.from_dataframe(
    df,
    output='fake_buoy.nc',
    attributes=attributes,
    axes={'t': 'time', 'x': 'longitude', 'y': 'latitude', 'z': 'depth'}
)
```

ocefpaf:
Thanks @kwilcox and @jbosch-noaa for the review! This should be ready to go now.

PS: @kwilcox I decided to use non-default axes and map them b/c while adding a note that this step is not needed if the axes are named `'t, y, x, z'`.

PR #307, opened 2018-04-16
environment without R pkgs

ocefpaf:
@rsignell-usgs we can remove even more from the environment if you really need a smaller env.


rsignell-usgs:
Thanks!

PR #308, opened 2018-04-25
Fix Windows installation of the default env

ocefpaf:
This should help users that are using the IOOS env as their default Python distribution to install it on Windows without any workarounds.
(The workaround is in Travis-CI now.)


ocefpaf:
for some reason the Boston Light Swim notebook fails on Travis-CI but passes locally :-/

PR #310, opened 2018-05-14
update compliance-checker notebook

ocefpaf:
@jbosch-noaa this updates the notebook to use `compliance-checker 4.0.0`. Should we re-publish it with a new name/date? I left the original "2017-02-10-running_compliance_checker.ipynb" here for now.


jbosch-noaa:
Yes. Thank you. I think we should re-publish with a new date.  We should also take the old one down since it is with an out of date version of the checker.

ocefpaf:
Self-merging this one b/c it is only an update.

PR #320, opened 2018-11-30
Fix notebooks testing

ocefpaf:
Hopefully make it faster and more reliable with new technologies like `nbval` and `pytest-xdist`.


ocefpaf:
Some of the failures here are due to https://github.com/Unidata/netcdf-c/issues/1299

ocefpaf:
This PR is becoming massive and there are many other notebooks to fix. I'll merge this one to avoid dealing with conflicts later and finish the other notebooks in another PR. (The fixes are mostly missing data, rotten links, and update to new software API like the change in how pandas handle tz.)

PR #305, opened 2018-03-21
IOOS models

ocefpaf:
# Not ready for merging yet

@jbosch-noaa here are the hypothes.is links:

### Grid and model statistics:

http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/ioos_models/notebooks/2018-04-30-grids.ipynb

### Horizontal slices for temperature:

http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/ioos_models/notebooks/2018-04-30-temperature-horizontal.ipynb

### Horizontal slices for salinity:

http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/ioos_models/notebooks/2018-04-30-salinity-horizontal.ipynb

### Horizontal slices for sea surface height:

http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/ioos_models/notebooks/2018-04-30-sea-surface-height-horizontal.ipynb

This notebook only deals with the various models grids at the moment. Once we define all the models we want and how to show the grids we can expand with horizontal and vertical slices of temperature, salinity, etc.

Closes #304, #284 

PS: check the models available on https://eds.ioos.us/

PR #321, opened 2018-12-02
IOOS models grids

ocefpaf:
@dpsnowden, @sblockhartzzero, @rsignell-usgs and @mwengren this is the first notebook, grid statistics, for a series of notebooks on the ocean models available in IOOS datasets. (See https://github.com/ioos/notebooks_demos/pull/305 for some background on these notebooks.)

There is no "text/story" yet, just a draft of the code for a preliminary review. Please let me know what you think and if this approach is OK with you.


ocefpaf:
@sblockhartzzero you can make HTML annotations using the link below:

https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/grids/notebooks/2018-12-04-grids.ipynb

sblockhartzzero:
Felipe,

I will review this today. I assume we are still on for a call tomorrow. Let
me know if that's not correct. (There is a federal holiday tomorrow, but
I'll be working from home.)

Thanks,
Steve

ocefpaf:
> There is a federal holiday tomorrow, but I'll be working from home

If it is better for you we can move it to Friday.
(I will be available tomorrow but I'm afraid the internet connection at my location will be a problem.)

mwengren:
How about Friday afternoon sometime after 2PM?  That would be best for me, but go ahead with normal time tomorrow if it's better for you both.  Tomorrow I will be doing things other than work.  It looks like we all have edit rights to the calendar item so let's move the meeting update discussion there.

ocefpaf:
@sblockhartzzero this is ready for review.

PS: the next notebook will be from the DMAC webinar

https://github.com/ocefpaf/2018-10-25-DMAC-TechWebinar

and the ones in https://github.com/ioos/notebooks_demos/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

ocefpaf:
@sblockhartzzero this one is ready for review.

PS: you can make HTML annotations using the link below.

[via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/grids/notebooks/2018-12-04-grids.ipynb](https://via.hypothes.is/http://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/grids/notebooks/2018-12-04-grids.ipynb)

ocefpaf:
@sblockhartzzero reported to me via e-mail that the PacIOOS model is reporting time step as `0`. That is a bug in the detection of the valid time coordinate (model run vs time). Fixing this in my next commit.

ocefpaf:
@sblockhartzzero I addressed your comments (e-mail conversation) and this is ready to for another round of review/merge.

mwengren:
@ocefpaf After updating my IOOS environment yesterday, I was able to run the notebook without trouble.  

I have a few comments I'll add in here.  I didn't want to go to the trouble to create an account in Hypothesis for just a few notes (they should enable GitHub OAuth IMO).

At the top, it would be nice to inclde a link to a GitHub repo for tools you mention in the comments (gridgeo for example).  That way people (like me) who may want to learn more can go directly there from the notebook if they want.  Maybe I'm being lazy, but after reading your description I wanted to just be able to link directly somewhere to find more info, but not quite Google for it.  You have it for GeoJSON I see.

Also, I think it might help to explain why a few of the cells (mostly the plotting ones) take so long to execute.  As an unfamiliar user the first time I ran the notebook, I found myself wondering if something had gone wrong with my setup after all, because often you have to wait quite a while.  Some ideas:

- cell 2: maybe just explain that the model outputs are being downloaded from the URLs specified, so there's an associated delay with that
- cell 4: this one is odd, the first time I ran it it was quite slow, but then even after restarting kernel and clearing outputs, running the full notebook a second time it was much faster.  What's going on there?
- cell 6: strange as well, first time slow, second time much faster, although for this one, after restarting kernel and rerunning all cells, it was still slower. (~ 2 - 3 mins, and my laptop is pretty fast)

In cell 6, it doesn't appear much is happening in the code, so a short explanation of why the execution time is so long might help (what exactly is the step that is taking the most time and what is it doing).  Is it the step to calculate the bounding polygons in Gridgeo, or is it the repeated call from cell 4 to:
```
df = get_stats(name, model)
html = to_html(df)
```
Come to think of it, could the output from cell 4 be reused here to save time? 

Or, what is this line doing?   `geojson = model['grid'].outline.__geo_interface__`.  It's not clear.

The notebook is great and it's really useful to see how to work the the model data in Python, I plan to share with a few people once it's done.  I just think a little more detail to explain the slowness and illuminate what's happening in some of the code blocks would help.

ocefpaf:
> At the top, it would be nice to inclde a link to a GitHub repo for tools you mention in the comments (gridgeo for example).

Indeed. I added a similar link to gridgeo docs.

> * cell 2: maybe just explain that the model outputs are being downloaded from the URLs specified, so there's an associated delay with that

Done. (We should do a "grab a coffee" warning b/c most of our notebooks fetch tons of data online and can be quite slow sometimes.)

> * cell 4: this one is odd, the first time I ran it it was quite slow, but then even after restarting kernel and clearing outputs, running the full notebook a second time it was much faster.  What's going on there?

Note sure. It should take the same amount of time when restarting the kernel. Nothing is saved or cached between kernel sessions. Without re-starting the kernel there is caching though and we expect it to be a little bit faster.

> * cell 6: strange as well, first time slow, second time much faster, although for this one, after restarting kernel and rerunning all cells, it was still slower. (~ 2 - 3 mins, and my laptop is pretty fast)
> In cell 6, it doesn't appear much is happening in the code, so a short explanation of why the execution time is so long might help (what exactly is the step that is taking the most time and what is it doing). Is it the step to calculate the bounding polygons in Gridgeo, or is it the repeated call from cell 4 to:

The line `geojson = model['grid'].outline.__geo_interface__` creates and caches the grid outline. It is slow b/c it relies on shapely to do so. It is cached if running again in the same session.

BTW, `get_stats(name, model)` is quite fast and not really and we are creating individual tables now, for the popups, instead of a single one.

> I just think a little more detail to explain the slowness and illuminate what's happening in some of the code blocks would help.

Hopefully the new commits will clarify that. Thanks for the review!

mwengren:
@ocefpaf Changes look good, thanks!

ocefpaf:
> @ocefpaf Changes look good, thanks!

Thanks! Self-merging this and moving to the next one.

PR #325, opened 2019-01-27
use non-R env as default

ocefpaf:
This PR updates the IOOS env and change the default env to be Python only while keeping an extra env file for those who need R. The idea is to reduce incompatibilities due to the slow development in R packages and let Python users take advantage of the latest package version.

There are also some environment simplifications via the removal of packages that would be pulled b/c they are dependencies of other packages, and upgrades the env to Python 3.7.


ocefpaf:
The environment on Windows seems broken but I believe https://github.com/conda-forge/ipykernel-feedstock/pull/43 will fix it. Merging this to get Python 3.7 IOOS env out.

PR #327, opened 2019-02-04
fix two more tz issues

ocefpaf:
Ping @rsignell-usgs


rsignell-usgs:
Let's fix the #fillmismatch too, or should that be a separate PR?

PR #331, opened 2019-02-25
simplify page publication

ocefpaf:
Publishing with `doctr` should be easier safer.


ocefpaf:
Let's test this on master!

PR #333, opened 2019-03-04
IOOS models part 2

ocefpaf:
This notebook is a continuation of #305.

@sblockhartzzero this is 50% ready (missing the text) here is the hypothes.is link in case you want to check the code:

https://via.hypothes.is/https://nbviewer.jupyter.org/github/ocefpaf/notebooks_demos/blob/grid_part2/notebooks/2019-03-08-grids-temperature.ipynb


ocefpaf:
The last commit follows up on this e-mail review from @sblockhartzzero:

---
Thanks for the review Stephen!
I just pushed some new commits addressing your comments.

Please let me know if that is working for you.
I have detailed comments below.
 
> I did get a DAP server error on WCOFS, but that's probably a temporary thing.

Indeed. I got the model in my run.

> Just a few issues.
> 1) In your comments, I think you should avoid words like wonky or flaky. You should definitely identify issues but I think be more diplomatic. If there are issues with the models, is there a way we could provide useful feedback to the model owners?

Fixed. Some of those were already reported upstream.

> 2) Cell 16 not getting recognized as a comment, gives error:
File "<ipython-input-16-2440e4bdda99>", line 1
    Last but not least we have `plot_surface_cfvar` that checks if we have a UGRID,
           ^
SyntaxError: invalid syntax
All comment cells should be markdown now. 
 
> 3) Got plots for all except WCOFS (which had the DAP error) and WestCoastUCSC (which comes after WCOFS on the list. Right before the plots, there was an error:
KeyError                                  Traceback (most recent call last)
<ipython-input-18-300d56d7c16b> in <module>
      1 for name in models.keys():
      2     model = models[name]
----> 3     grid = model['grid']
      4     var = model['variable']
      5     vlevel = get_surface_idx(var, grid.mesh)

> KeyError: 'grid'

I added a `key` drop to account for that case.

> So, it's understandable that WCOFS model didn't plot, but not sure why UCSC model didn't plot.

It should be fixed with they drop.

ocefpaf:
@sblockhartzzero based on your last e-mail it seems that this one is good to go, right? Can you merge it?

sblockhartzzero:
Done

ocefpaf:
> Done

Thanks! Part three is underway!!

PR #338, opened 2019-05-28
use pangeo binder

ocefpaf:
@rsignell-usgs you are more updated with pangeo than me. Do you think this move is OK? Or should we not be using the pangeo binder until it is out of beta?


ocefpaf:
After asking upstream I decided to go with this b/c, in a way, tradicional binder is also not that stable. Another reason is that we need to raise awareness among our users to the technologies we are relying on, like pangeo.

Issue #339, opened 2019-05-30
Pendulum error on Irma notebook


rsignell-usgs:
I'm getting an error when running the Hurricane Irma notebook on binder. 

Cell [8] is producing:
```python-traceback
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-8-5ab3fb494fb2> in <module>
      7 end = points[-1]['FLDATELBL']
      8 
----> 9 start = pendulum.from_format(start, fmt='YYYY-MM-DD h:mm A ddd').naive()
     10 end = pendulum.from_format(end, fmt='YYYY-MM-DD h:mm A ddd').naive()

/srv/conda/envs/notebook/lib/python3.7/site-packages/pendulum/__init__.py in from_format(string, fmt, tz, locale)
    254     Creates a DateTime instance from a specific format.
    255     """
--> 256     parts = _formatter.parse(string, fmt, now(), locale=locale)
    257     if parts["tz"] is None:
    258         parts["tz"] = tz

/srv/conda/envs/notebook/lib/python3.7/site-packages/pendulum/__init__.py in now(tz)
    213 
    214     if tz is None or tz == "local":
--> 215         dt = _datetime.datetime.now(local_timezone())
    216     elif tz is UTC or tz == "UTC":
    217         dt = _datetime.datetime.now(UTC)

/srv/conda/envs/notebook/lib/python3.7/site-packages/pendulum/tz/__init__.py in local_timezone()
     57     Return the local timezone.
     58     """
---> 59     return get_local_timezone()

/srv/conda/envs/notebook/lib/python3.7/site-packages/pendulum/tz/local_timezone.py in get_local_timezone()
     29 
     30     if _local_timezone is None:
---> 31         tz = _get_system_timezone()
     32 
     33         _local_timezone = tz

/srv/conda/envs/notebook/lib/python3.7/site-packages/pendulum/tz/local_timezone.py in _get_system_timezone()
     57         return _get_darwin_timezone()
     58 
---> 59     return _get_unix_timezone()
     60 
     61 

/srv/conda/envs/notebook/lib/python3.7/site-packages/pendulum/tz/local_timezone.py in _get_unix_timezone(_root)
    236         return TimezoneFile(tzpath)
    237 
--> 238     raise RuntimeError("Unable to find any timezone configuration")
    239 
    240 

RuntimeError: Unable to find any timezone configuration
```
Here is my `conda list`:
<details>
[...]

PR #340, opened 2019-05-30
Use only standard library datetime instead of pendulum

ocefpaf:
Closes https://github.com/ioos/notebooks_demos/issues/339


ocefpaf:
The failures here are b/c http://tds.marine.rutgers.edu/thredds/ is down. I'll restart the CIs when the service is up again.

rsignell-usgs:
Looks good to me!

PR #342, opened 2019-06-25
test miniconda rc with 4.7

ocefpaf:
Ping @msarahan.


msarahan:
Waiting with bated breath. Come on green....

ocefpaf:
The failure is unrelated, so I guess it is working.
I need to split the actual env creation as its own test and then the notebooks into another.

ocefpaf:
@msarahan all rcs of miniconda are building the IOOS environment without a problem. (I also send the links to some people to test. No one report problems.)

msarahan:
Thanks for the testing!

ocefpaf:
BTW, if you set up a "nightly" URL for miniconda we could use that to continuously test it here.

msarahan:
Will do. We have them in ci, but we need to publish them.

PR #347, opened 2020-02-13
Find IOOS models catalog

ocefpaf:
Not ready for merging yet. Needs some tweaking to actually find data.

### DIFF
--- a/notebooks/2020-02-28-IOOS-models.ipynb
+++ b/notebooks/2020-02-28-IOOS-models.ipynb
@@ -1,0 +1,209 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# IOOS models: how to find model data in the catalog\n",
+    "\n",
+    "This is the third post on the series “IOOS Ocean Models IOOS.”"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "code_folding": []
+   },
+   "outputs": [],
+   "source": [
+    "models = {\n",
+    "    \"DOPPIO\": {\n",
+    "        \"RA\": \"MARACOOS\",\n",
+    "        \"url\": \"http://tds.marine.rutgers.edu/thredds/dodsC/roms/doppio/2017_da/avg/Averages_Best\",\n",
+    "        \"var\": {\"standard_name\": \"sea_water_potential_temperature\"},\n",
+    "    },\n",
+    "    \"NYHOPS\": {\n",
+    "        \"RA\": \"MARACOOS\",\n",
+    "        \"url\": \"http://colossus.dl.stevens-tech.edu:8080/thredds/dodsC/latest/Complete_gcmplt.nc\",\n",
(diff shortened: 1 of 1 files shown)


ocefpaf:
@kbailey-noaa and @MathewBiddle the metadata that would make this notebook [possible is coverage_content_type](https://wiki.esipfed.org/Attribute_Convention_for_Data_Discovery_1-3#coverage_content_type):


**coverage_content_type**: An ISO 19115-1 code to indicate the source of the data (image, thematicClassification, physicalMeasurement, auxiliaryInformation, qualityInformation, referenceInformation, **modelResult**, or coordinate).

MathewBiddle:
I wonder how many are using it. 

* [ACDD highly recommends](https://wiki.esipfed.org/Attribute_Convention_for_Data_Discovery_1-3#Index_by_Attribute_Name) it. 
* [NCEI has it listed](https://www.nodc.noaa.gov/data/formats/netcdf/v2.0/)
* [IOOS Metadata Profile v1.2](https://ioos.github.io/ioos-metadata/ioos-metadata-profile-v1-2.html) doesn't explicitly call it out.

It looks like only [one example in the gold standard ERDDAP](https://standards.sensors.ioos.us/erddap/search/index.html?page=1&itemsPerPage=1000&searchFor=coverage_content_type) is using it. But nothing else.

mwengren:
I recall a related discussion of this topic in Catalog years ago.  It was also related to the `source` attribute, which [CF indicates](https://cfconventions.org/cf-conventions/cf-conventions.html#description-of-file-contents) should name the numerical model used if data was produced by a model. 

Keep in mind that metadata that goes into the IOOS Catalog is first passed through ISO XML produced by ncISO (for THREDDS) or by ERDDAP directly, before it can be indexed by Catalog.  So unless ncISO or ERDDAP includes `source` or `coverage_content_type` in the resulting ISO XML, Catalog won't know about it.

At one point I tried to understand how ncISO was treating the `source` attribute, and it turned out it wasn't reading it at all (see Unidata/threddsIso#34).   Not sure how that issue was ultimately resolved if at all with the two different ncISO versions.

I haven't searched for `coverage_content_type` in Catalog datasets, but it looks like it is read/used in this XSL stylesheet: [UnidataDD2MI.xml](https://github.com/Unidata/threddsIso/blob/master/src/main/resources/xsl/nciso/UnidataDD2MI.xsl) so should be available if present.

Finally, on a related note, we recently [updated Catalog](https://github.com/ioos/catalog-ckan/issues/208) to read IOOS Profile 1.2 attributes directly from ERDDAP datasets, since this was easier than trying to adapt the ERDDAP ISO XML output.  This could be enhanced to include either `source` or `coverage_content_type` if necessary, but it won't help much here since most model datasets will be served via the THREDDS/ncISO pathway. 

Also, the client code would need to be updated to read the CKAN API via [ckanapi](https://github.com/ckan/ckanapi) module rather than CS-W, because the CS-W records will not include these custom ERDDAP attributes.   

If we decide it would help, we can look into extending Catalog functionality to read THREDDS attributes as well.

cc @benjwadams.

Issue #355, opened 2020-10-23
Error from the first line in the using r-obistools and r-obis to explore the OBIS database notebook


MathewBiddle:
The [obis notebook](https://github.com/ioos/notebooks_demos/blob/master/notebooks/2018-02-20-obis.ipynb) fails on the first line when running locally through Jupyter notebook in the IOOS environment using Python 3 Kernel:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-fb23c6edefe4> in <module>
----> 1 get_ipython().run_line_magic('load_ext', 'rpy2.ipython')

~\programs\Anaconda3\envs\IOOS\lib\site-packages\IPython\core\interactiveshell.py in run_line_magic(self, magic_name, line, _stack_depth)
   2324                 kwargs['local_ns'] = self.get_local_scope(stack_depth)
   2325             with self.builtin_trap:
-> 2326                 result = fn(*args, **kwargs)
   2327             return result
   2328 

<decorator-gen-58> in load_ext(self, module_str)

~\programs\Anaconda3\envs\IOOS\lib\site-packages\IPython\core\magic.py in <lambda>(f, *a, **k)
    185     # but it's overkill for just that one bit of state.
    186     def magic_deco(arg):
--> 187         call = lambda f, *a, **k: f(*a, **k)
    188 
    189         if callable(arg):

~\programs\Anaconda3\envs\IOOS\lib\site-packages\IPython\core\magics\extension.py in load_ext(self, module_str)
     31         if not module_str:
     32             raise UsageError('Missing module name.')
---> 33         res = self.shell.extension_manager.load_extension(module_str)
     34 
     35         if res == 'already loaded':

~\programs\Anaconda3\envs\IOOS\lib\site-packages\IPython\core\extensions.py in load_extension(self, module_str)
     78             if module_str not in sys.modules:
     79                 with prepended_to_syspath(self.ipython_extension_dir):
---> 80                     mod = import_module(module_str)
     81                     if mod.__file__.startswith(self.ipython_extension_dir):
     82                         print(("Loading extensions from {dir} is deprecated. "

~\programs\Anaconda3\envs\IOOS\lib\importlib\__init__.py in import_module(name, package)
    125                 break
    126             level += 1
--> 127     return _bootstrap._gcd_import(name[level:], package, level)
    128 
    129 

[...]

PR #358, opened 2020-10-28
use conda-lock on CIs

ocefpaf:
This should reduce the time it takes to install packages by cutting on the solver resolution.


ocefpaf:
Next step would be to setup out binder with the lockfile.
That would take care of #355

PR #359, opened 2020-10-29
fix publication with latest nbconvert

ocefpaf:
New versions of nbconvert need to pass the template file.


ocefpaf:
Travis-CI hook is not showing up b/c we need to migrate to travis.com. This repository is still on travis.org.

Issue #248, opened 2017-07-14
Create an R Notebook to align data to Darwin Core


jbosch-noaa:
This is a GitHub issue to track my converting Abby Benson's R script to align data to Darwin Core for submission to OBIS.

ocefpaf:
> Abby Benson's R script to align data to Darwin Core for submission to OBIS.

@jbosch-noaa do you have that script online somewhere?

jbosch-noaa:
@ocefpaf - No. I have it in an email I just forwarded to you.

MathewBiddle:
See [Abby's R script](https://github.com/ioos/bio_data_guide/tree/master/Standardizing%20Marine%20Biological%20Data/datasets/example_script_with_fake_data). I made a [Python version](https://github.com/ioos/bio_data_guide/pull/40) and am waiting for it to get merged.

MathewBiddle:
Python notebook is available [here](https://github.com/ioos/bio_data_guide/blob/master/Standardizing%20Marine%20Biological%20Data/datasets/example_script_with_fake_data/IOOS%20DMAC%20DataToDwC_Notebook_event.ipynb). It needs more context if we would like to add it to here.

ocefpaf:
> Python notebook is available [here](https://github.com/ioos/bio_data_guide/blob/master/Standardizing%20Marine%20Biological%20Data/datasets/example_script_with_fake_data/IOOS%20DMAC%20DataToDwC_Notebook_event.ipynb). It needs more context if we would like to add it to here.

Do you want to send a draft PR or should I? We can work on the text/context later.

MathewBiddle:
I can. It'll be good to freshen up my git skills.

MathewBiddle:
Do we still want an R version of this, or does the python version suffice?

albenson-usgs:
An R version [already exists](https://github.com/ioos/bio_data_guide/blob/master/Standardizing%20Marine%20Biological%20Data/datasets/example_script_with_fake_data/IOOS_DMAC_DataToDwC_Notebook_event.md) seems like it wouldn't be that hard to have both so people can use the language they are more familiar with. Would be great if there was some way to display it in the Code Gallery side by side so people could select the one most appropriate to them. Maybe that's a feature beyond the Code Gallery's functionality though.

MathewBiddle:
We do have that R version linked at the bottom of this notebook. 
[...]

PR #371, opened 2020-12-08
Adding data conversion to DwC notebook

MathewBiddle:
Added the source data file to data/. Commented out the bits where we write the resultant data file. Not sure where we'd want to write them, or if we want to? 

Addresses #248


MathewBiddle:
@ocefpaf, I think the most recent pull request lays out the process a little better and provides links to other resources. If you don't mind taking a look, I'd appreciate any feedback.

ocefpaf:
I'm adding pyworms to conda-forge and then to our env so we can run this. Be back here to review as soon as I finish that...

ocefpaf:
This notebook will need pyworms. I'm adding it in 

> I'm adding pyworms to conda-forge and then to our env so we can run this. Be back here to review as soon as I finish that...

Pending: https://github.com/iobis/pyworms/issues/2

MathewBiddle:
Feedback from Abby:
   1. The text looks good. The only thing you need to change is this line: 
       ```event['eventDate'] = pd.to_datetime(event['date'],format='%m/%d/%Y',utc=True)```
      * You need to remove the `utc=True` part. There is no time information included in the file so we should not add it. Having YYYY-MM-DD is sufficient and is the only information that's recorded. By adding time, you're creating data that's not there and is not accurate.
   1. adjust `writting` to `writing`.
   1.  The way you described what the eventID is doing is missing some information. You say "For this we need to combine all of the measurements or facts that we want to include with the data package into one table using the eventID as a connection to the occurrence records." but it also links to the mof file and not to the occurrences. So for instance this measurement 

0 | St. John_250_1 |   | temperature | http://vocab.nerc.ac.uk/collection/P25/current... | 25.2 | Celsius | http://vocab.nerc.ac.uk/collection/P06/current... | 3 | 2004-07-16 00:00:00+00:00
-- | -- | -- | -- | -- | -- | -- | -- | -- | --

only has an eventID and does not have an occurrenceID so in this instance the eventID is linking the measurement (temperature) to the event but it's not linking to the occurrence.

ocefpaf:
LGTM. Do you want to remove the commended out code cells before we merge?
Also, pyworms is not on conda-forge b/c of https://github.com/iobis/pyworms/issues/2
Hopefully we can solve that one with upstream and package it.

MathewBiddle:
Yeah, I'll uncomment those lines. 

Should we have the output files written to a subdir in `data/` ? Might be nice to have them available for folks to review.

Should I rebase and merge?

ocefpaf:
> Should we have the output files written to a subdir in `data/` ? Might be nice to have them available for folks to review.

That would be nice. Specially for folks who will try to run this locally.

> Should I rebase and merge?

No need. Without pyworms that won't help much.

MathewBiddle:
@ocefpaf, I created a new data structure for these data:
```
data
└───dwc
    ├───processed
    │       MadeUpData_event.csv
    │       MadeUpData_mof.csv
    │       MadeUpData_Occurrence.csv
    │
    └───raw
            MadeUpDataForBiologicalDataTraining.csv
```

If you're good, feel free to merge.

ocefpaf:
> If you're good, feel free to merge.

Do we want this without pyworms? We cannot install it in our envs until https://github.com/iobis/pyworms/issues/2 is solved. Maybe we should ping the author again.

MathewBiddle:
can we PR a license to that repo?

MathewBiddle:
can we push this out without pyworms (I do include a note on how to install via github and pip), and work with them on the license?

ocefpaf:
> can we PR a license to that repo?

We can but that is something the author should do.

> can we push this out without pyworms (I do include a note on how to install via github and pip), and work with them on the license?

We would need to skip it from the CIs, but sure, that is a possibility. However, ideally we should not encourage use of unlicensed code IMO.

ocefpaf:
@MathewBiddle we have pyworms on conda-forge! Are you done with this PR? Should we merge it?

MathewBiddle:
@ocefpaf Wahoo! If you've taken a look and think it's good, then let's merge it!

MathewBiddle:
Actually, I need to remove the git installation note since it's available on conda-forge.

MathewBiddle:
Okay, now it should be good to merge.

MathewBiddle:
@ocefpaf looks like we built the urls to the files incorrectly. The notebook links to https://github.com/master/notebooks_demos/blob/master/notebooks/data/dwc/processed/MadeUpData_Occurrence.csv
when it should be
https://github.com/ioos/notebooks_demos/blob/master/notebooks/data/dwc/processed/MadeUpData_Occurrence.csv

PR #380, opened 2021-01-31
build and deploy docs via GHA

ocefpaf:
We are still not deploying, so let's delay merging this until we can find a GHA that handles that or when https://github.com/drdoctr/doctr/pull/372 is complete.


ocefpaf:
Let's test if this will run on master, if not I'll try the ssh key route, is yes we'll have to skip the deployment on PRs.

PR #384, opened 2021-02-01
Let's try to implement a binder image cache

ocefpaf:
This should not only help our binder instance to start up quickly but also make RStudio available.


ocefpaf:
I need to re-do this with newer tech that appeared while we were writing this.

PR #397, opened 2021-03-12
update env

ocefpaf:
r-obistools was missing from the env and I also re-run that notebook to ensure everything is working.


ocefpaf:
The failure is unrelated to the change. Merging it...

PR #372, opened 2021-01-11
Extend time in gts notebook

MathewBiddle:
I've made the time constraint more dynamic. Since the `now` functionality [doesn't work with erddapy](https://github.com/ioos/erddapy/issues/164), I've done it using timedelta.

Issue #351, opened 2020-04-20
New GTS counts graphic (notebook...?)


kbailey-noaa:
For @ocefpaf 
I'm not actually sure if a notebook is what I'm looking for, but figured you could advise:
The goal is to embed a new graphic in one of our ioos.noaa.gov webpages - probably [here](https://ioos.noaa.gov/data/contribute-data/provision-data-gts/) - that shows volume of IOOS data delivered to the GTS per month.  The graphic would automatically update monthly.

We have statistics (.csv files generated monthly) gathered by NDBC [here](http://www.ndbc.noaa.gov/ioosstats/) that Kevin O'Brien has already put into an ERDDAP here:
https://ferret.pmel.noaa.gov/generic/erddap/tabledap/ioos_obs_counts.html
He downloaded files from Jan 2019 forward.  You can search on platform ID, or sponsor or region.  He also provided me the configuration file in case we want to set up the ERDDAP elsewhere.

I'm looking for a simple bar graph that shows total # obs per month going to the GTS:
![GTSflow](https://user-images.githubusercontent.com/11823385/79789671-9033f780-8318-11ea-8031-cd016a807593.png)

The # obs is a grand total of all stations, wave + met obs.

ocefpaf:
I'll write something as a notebook that we can use to update the page later. Figuring out how to automate this is the trick part though.

kbailey-noaa:
@ocefpaf I've asked @MathewBiddle to help with this since I've been a bit overwhelmed lately.  Matt joined on Monday and part of his duties include metrics/outreach related to our products, so this fits in well.  Matt - See the pull request for the preview.  We can discuss in more detail outside of github. 
Source files are here: https://www.ndbc.noaa.gov/ioosstats/
and the ERDDAP Kevin O stood up is here: https://ferret.pmel.noaa.gov/generic/erddap/tabledap/ioos_obs_counts.html
Met = winds, air pressure, air temp, dew point, water temp
Wave = wave height and dominant wave period
IOOS-regional = RAs plus their affiliates or parters (regardless of funding relationship) as accounted for on the [NDBC website](https://www.ndbc.noaa.gov/to_station.shtml)
Filipe - Is this issue supposed to be closed/replaced with the pull request? Unclear if I should've added this comment to the pull request instead.

ocefpaf:
[...]

PR #398, opened 2021-03-12
creating widgets for GTS notebook.

MathewBiddle:
fixes https://github.com/ioos/notebooks_demos/pull/372 and continues #351


MathewBiddle:
FYI, I updated the notebook to pull directly from NDBC. I also added [google colab](https://colab.research.google.com/github/MathewBiddle/notebooks_demos/blob/GTS_widgets/notebooks/2020-10-10-GTS.ipynb) link if you'd like to use that. Should work out of the box now (other then the mapping).

MathewBiddle:
Only have data from 2018-01 to the month before the date you run the script (eg. today only has data up to 2021-03). Need to do some checks or limit the date range.

Issue #417, opened 2021-07-02
Jupyter{Book} Enhacement


lohithmunakala:
- [x] Add a standardized, centered IOOS header/masthead logo that links back to ioos.us
- [x] Make the IOOS image on the upper left point to ioos.us (we'll need to add the home link back into the menu probably)
- [x] footer: standardize to ioos.us design, add link to ioos.noaa.gov, contact info etc. Should have consistent look and feel if possible and match the content at the main ioos.us page

lohithmunakala:
![Screenshot from 2021-07-07 01-05-08](https://user-images.githubusercontent.com/56470584/124656898-6b955280-debf-11eb-94f1-ac70f0cfa199.png)


@ocefpaf @MathewBiddle is this what we are looking at for the footer section? Links to respective social media handles and mail to webmaster.ioos.us@noaa.gov (given on the ioos.us website)? 

For the logo, should we change it or keep it the same? There isn't an option to link it back to ioos.us btw. I have added the link to ioos.us below in the footer section of the menu.

MathewBiddle:
I guess it checks the boxes. 

The footer on ioos.us looks like this:
![image](https://user-images.githubusercontent.com/8480023/124661629-be4a1c80-de75-11eb-8471-a5b0c06b2d0e.png)
I don't think we will be able to replicate that fancy look through the JB.

@mwengren what do you think about the data demo center having the `Contact us` link go to webmaster?

I wonder if it would look better to make the logo larger? It just seems like a lot of text on that upper left portion.

Any other opinions are welcome.

mwengren:
Hmm... that's a toss up.  All the ioos.us pages Contact Us footer links send email to webmaster.ioos.us@noaa.gov I think. 

But on our documentation pages at ioos.github.io, we also have a Feedback link at the top of the nav bar that sends email to data.ioos@noaa.gov.  

I'd say if this is the only contact email on the JupyterBook pages, we should use data.ioos@noaa.gov - will be more direct to someone in Ops for any feedback we get.

MathewBiddle:
Agreed. Let's use data.ioos@noaa.gov. 

As for the logo, if you can increase the size that might be nice. Otherwise this looks good to me. 

Can we make the copyright look like ioos.us:
![image](https://user-images.githubusercontent.com/8480023/124798712-afb93f00-df21-11eb-847f-6732cbc9369c.png)

lohithmunakala:
[...]

Issue #413, opened 2021-06-28
New name for the IOOS Data Demo Center


MathewBiddle:
@mwengren @kbailey-noaa @ocefpaf 

Time to get creative! :)

I know we've discussed a rename to "Notebook Gallery". But, now we are moving towards using [Jupyter{book}](https://jupyterbook.org/intro.html) as the wrapper around the notebooks. So, I'm not sure if "Notebook Gallery" is still appropriate with that technology.

Looking through the Jupyter{book} [gallery of books](https://executablebooks.org/en/latest/gallery.html) the titles are slightly more descriptive. 

A couple of (not creative) thoughts: 
* "IOOS tools for ocean observations"
* "IOOS code suite"
* "Ocean observing system notebooks"

At a minimum, we can pick one of "Notebook Gallery" or "IOOS Data Demo Center" but we need to be consistent in our text. I see various names listed throughout [the site](https://ioos.github.io/notebooks_demos/) now.

mwengren:
'IOOS Code Lab', 'IOOS CodeLab', 'IOOS Code Gallery'?  If we want to get away from the Notebook term.

You could also go with  Data Lab DataLab instead I suppose.  I'd vote against 'IOOS Code & Data Lab' just because that is too cumbersome.  

Just a few ideas.

ocefpaf:
I like `IOOS Code Lab`.

MathewBiddle:
I like `IOOS CodeLab` . When you [search for "codelab"](https://www.google.com/search?q=codelab&client=firefox-b-1-d&sxsrf=ALeKk02I6HfH3JPORvyocYB0skbobH5vOg%3A1624986239533&ei=f1LbYKmDINTS5NoPyqmt2AY&oq=codelab&gs_lcp=Cgdnd3Mtd2l6EAMyCggAELEDEIMBEEMyAggAMgYIABAHEB4yAggAMgYIABAHEB4yAggAMgIIADICCAAyAggAMgIIADoHCAAQRxCwAzoHCAAQsAMQQ0oECEEYAFD5pwRY-acEYJKrBGgBcAJ4AIABZ4gBwwGSAQMxLjGYAQCgAQGqAQdnd3Mtd2l6yAEKwAEB&sclient=gws-wiz&ved=0ahUKEwjpvsm8qb3xAhVUKVkFHcpUC2sQ4dUDCA4&uact=5) a few similar sites come up.

MathewBiddle:
[...]

PR #423, opened 2021-07-07
Jupyter{Book} Enhancement 

lohithmunakala:
This enhancement is based on the https://github.com/ioos/notebooks_demos/issues/417 (issue). 

A new footer section has been added and the logo has been made bigger.

### The homepage looks like this now
image


MathewBiddle:
Just pulled this over and built locally. Looks like the colab link has something wrong.
If I try to launch Colab, as shown in the screen shot below, I get a link to the local notebook not a colab notebook.
image

Here's the link I get: 
`file:///github/ioos/notebooks_demos/blob/master/jupyterbook/content/Code%20Gallery/data_access_notebooks/2016-10-12-fetching_data.ipynb`

I think _config.yml needs to be updated, https://github.com/lohithmunakala/notebooks_demos/blob/36c6bcb7cb1cb93392d65599fbbf5391110511f2/jupyterbook/_config.yml#L77

jupyterbook/_config.yml
```diff
@@ -40,13 +40,21 @@ parse:
...
 html:
-  favicon                   : "images/IOOS-logo_little.png"  # A path to a favicon image
+  favicon                   : "images/IOOS_logo.png"  # A path to a favicon image
   use_edit_page_button      : false  # Whether to add an "edit this page" button to pages. If `true`, repository information in repository: must be filled in
   use_repository_button     : false  # Whether to add a link to your repository button
   use_issues_button         : true  # Whether to add an "open an issue" button
-  extra_navbar              : Powered by <a href="https://jupyterbook.org">Jupyter Book</a>  # Will be displayed underneath the left navbar.
-  extra_footer              : ""  # Will be displayed underneath the footer.
-  google_analytics_id       : ""  # A GA id that can be used to track book views.
+  extra_navbar              : <a href="https://ioos.us">IOOS</a>   # Will be displayed underneath the left navbar.
```
MathewBiddle, line 47: +1 I like this addition!

MathewBiddle:
@mwengren see the screen shot in the [first comment](https://github.com/ioos/notebooks_demos/pull/423#issue-685429365). Is that footer sufficient for alignment with ioos.us sites?

IMHO it's not very pretty, but it relays the information you requested. I think just having the ioos.noaa.gov, Facebook, Twitter, and Contact Us links would suffice.

mwengren:
Looks good to me.  We should probably remove the Copyright though as I don't think that's the intention and everything we produce should be public domain.  We may have that elsewhere in our ioos.github.io sites as well and it should be removed there as well if so (separate issue).

MathewBiddle:
@lohithmunakala can we remove the copyright?

lohithmunakala:
> Just pulled this over and built locally. Looks like the colab link has something wrong.
> If I try to launch Colab, as shown in the screen shot below, I get a link to the local notebook not a colab notebook.
> image
> 
> Here's the link I get:
> `file:///github/ioos/notebooks_demos/blob/master/jupyterbook/content/Code%20Gallery/data_access_notebooks/2016-10-12-fetching_data.ipynb`
> 
> I think _config.yml needs to be updated, https://github.com/lohithmunakala/notebooks_demos/blob/36c6bcb7cb1cb93392d65599fbbf5391110511f2/jupyterbook/_config.yml#L77

I didn't add the Colab link in the `_config.yml` as weren't sure - initially - if we were going to need it. Just added it in this - https://github.com/ioos/notebooks_demos/pull/423/commits/15186f06d48483e3b32eef400c59bd3e931c5e3a - commit.  The option will now direct you to Colab.

lohithmunakala:
> @lohithmunakala can we remove the copyright?

There isn't a way we can remove the term "Copyright". This is what gets displayed when we remove anything from the copyright string in the `config` file. 

image

Or we can do something like this, which is the default

image

Which would we better?

MathewBiddle:
There might be a workaround, see this https://github.com/executablebooks/jupyter-book/issues/582#issuecomment-629371516

MathewBiddle:
@mwengren is it a deal breaker if we can't remove the copyright? Looks like it could get messy.

lohithmunakala:
> There might be a workaround, see this [executablebooks/jupyter-book#582 (comment)](https://github.com/executablebooks/jupyter-book/issues/582#issuecomment-629371516)

I did check this out, but as the static files that are generated are after we run the notebook, it would mess with the automatic publication of the website.  Because we use GitHub Actions to build the notebook, this would get overwritten every time we make a change to the repo - triggering the GA to create a new set of static files. 

We would probably need to automate editing out each of the CSS files and then publish the web page.

mwengren:
Ok, if it's too cumbersome to remove the Copyright, then we can leave as is.  It seems like the JupyterBook library should allow the option to customize that wording though, might be a good PR to submit to them if you know how it could be done.

jupyterbook/_config.yml
```diff
@@ -40,13 +40,21 @@ parse:
 #######################################################################################
 # HTML-specific settings
 html:
-  favicon                   : "images/IOOS-logo_little.png"  # A path to a favicon image
+  favicon                   : "images/IOOS_logo.png"  # A path to a favicon image
   use_edit_page_button      : false  # Whether to add an "edit this page" button to pages. If `true`, repository information in repository: must be filled in
   use_repository_button     : false  # Whether to add a link to your repository button
```
MathewBiddle, line 45: Let's turn these on too.
image

jupyterbook/_config.yml
```diff
@@ -40,13 +40,21 @@ parse:
...
   use_issues_button         : true  # Whether to add an "open an issue" button
```
MathewBiddle, line 46: Any way we can use the [existing template](https://github.com/ioos/notebooks_demos/blob/master/.github/ISSUE_TEMPLATE.md)?

MathewBiddle (commented):
I would like to have the GH edit and repo buttons available and maybe use our existing ticket template for the issues? Or, leave the defaults here and link somewhere to that ticket template.

jupyterbook/_config.yml
lohithmunakala: Sure!

jupyterbook/_config.yml
lohithmunakala: I'll check.

MathewBiddle:
@ocefpaf once those changes are made, I'm okay with merging. Unless you want to compile #413 in here and update the various references to `Notebook gallery` or `Data Demo Center` to `IOOS Code Lab`.

MathewBiddle:
@lohithmunakala can we rename this markdown to NANOOS.md

https://github.com/lohithmunakala/notebooks_demos/blob/jupyterbook_enhancement/jupyterbook/content/Video_Demo_Markdown/noos.md

lohithmunakala:
The header for the file inside? Or do we want the name of the file to be changed too.

MathewBiddle:
both

MathewBiddle:
Looks good! I'll merge.

PR #430, opened 2021-07-29
updating the cover image

MathewBiddle:
Adding new image and updating the intro.md to deal with transparent backgrounds.


ocefpaf:
The final art is really nice!

PR #431, opened 2021-08-02
Changing GHA to peaceiris

lohithmunakala:
Have updated the GHA in `config.yml` to peaceiris such that the webpage gets built automatically.


jupyterbook/_config.yml
```diff
@@ -1,7 +1,7 @@
 # Book settings
 title            : The U.S. Integrated Ocean Observing System (IOOS)  # The title of the book. Will be placed in the left navbar.
 author           : IOOS   # The author of the book
-copyright        : "2021"  # Copyright year to be placed in the footer
+copyright        : "2021"   # Copyright year to be placed in the footer
```
ocefpaf, line 4: @lohithmunakala just a reminder to open an issue upstream about this.

PR #437, opened 2021-08-13
Remove copyright

ocefpaf:
@MathewBiddle we can also add custom licenses if we need to with this config option.


MathewBiddle:
@ocefpaf was this an upstream update?

ocefpaf:
> @ocefpaf was this an upstream update?

Nope. I just read the docs :smile:

PR #440, opened 2021-08-18
adding arcpy examples

MathewBiddle:
starting arcpy examples. Don't know if we can install arcpy w/out ArcGIS Pro...


dpsnowden:
@MathewBiddle NOS has a site license for ArcGIS Pro so you should be able to install it if needed.  Recommend talking to Kathy or Tiffany if you need assistance.

